PR #14, opened 2020-10-23
Implemented config parsing for Chunk server

### LINKED ISSUE #4: Start-up configuration - Chunk server daemon
As an admin, I should be able to configure default options for the Chunk server daemon in a configuration file, and override them over the command line

abmono93:


Makefile
```diff
@@ -57,4 +67,4 @@ clean:
 mkdirs:
 	@-mkdir -p $(BUILD_DIRS:%=$(BUILD)/%)
 
-.PHONY: clean mkdirs all sadmd-bootstrap
+.PHONY: clean mkdirs all sadmd-bootstrap sadcd-bootstrap
```
damanmkumar: sadfsd-bootstrap should be added here as well

src/sadcd/sadcd.cpp
```diff
@@ -0,0 +1,8 @@
+#include <iostream>
```
damanmkumar: Isn't it unnecessary to have code that doesn't do anything? We can introduce these files when we write code that does something.

src/sadfsd/sadfsd-bootstrap.cpp
```diff
@@ -0,0 +1,3 @@
+int
```
damanmkumar: Same as sadcd.cpp

Makefile
abmono93: Right thanks good catch

damanmkumar (changes requested):
* Missing entry under `.PHONY` rule
* Files without any meaningful code need not be introduced now. `make` will fail, but that's OK since we can run `make sadmd-bootstrap`.

src/sadcd/sadcd.cpp
abmono93: Yes you're right I made thistrying to have something for bootstrap to exec to but since that isn't working yet it shouldn't be in this PR

damanmkumar (approved):

PR #18, opened 2020-10-28
implemented listener and socket classes

### LINKED ISSUE #12: Introduce class to listen for new connections
Introduce class to easily create sockets for listening on a particular port

### LINKED ISSUE #15: implement listener and socket classes without exceptions
part of user story - #12 
* exception support will be implemented with another task

damanmkumar:
**NOTE:** To be squashed and merged instead of plain merge

* can create a listener class by specifying ip and port number
* can accept connections
* ~yet to implement exception throwing when a class-invariant cannot be established.~
* exception support now implemented
* added example code to manually and minimally test `sadfs::inet::listener` and `sadfs::socket`

implements #15, and its parent #12

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,7 @@
+BIN_DIR = bin
+BUILD_DIRS = sadmd sadcd sadfsd comm example $(BIN_DIR)
 BUILD = build
-BUILD_DIRS = sadmd sadcd sadfsd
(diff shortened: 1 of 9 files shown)


abmono93 (approved):
Confirmed that it compiles on my Mac. And I was able to send messages from bin/msg to bin/echod

PR #19, opened 2020-10-28
sadcd-bootstrap starts chunk server daemon

### LINKED ISSUE #7: Bootstrap - Chunk server daemon
As an admin, I should be able to start the Chunk server daemon

abmono93:
closes #7

### DIFF
--- a/src/sadcd/sadcd-bootstrap.cpp
+++ b/src/sadcd/sadcd-bootstrap.cpp
@@ -97,21 +97,28 @@ display_help(po::options_description const& options)
 		options << '\n';
 }
 
+// Populates command-line args for sadcd and starts it
 void
 start_server(po::variables_map const& config)
 {
+	// Copy port from uint16 to C string
 	char port_str[2];
 	if (config.count("port"))
 	{
 		memcpy(port_str, &(config["port"].as<std::uint16_t>()), 2);
 	}else{
 		std::cerr << "Error: No port number was set\n";
+		std::exit(1);
 	}
+	
+	// Arguments needed for execve
 	const char *chunk_daemon = "sadcd";
 	char *const chunk_argv[] = {port_str, NULL};
 
 	errno = 0;
 	execve(chunk_daemon, chunk_argv, NULL);
+
+	// Execve only returns if an error occured
 	std::cerr << "Error: could not start chunk server: " << 
 		std::strerror(errno) << std::endl;
 	std::exit(1);
--- a/src/sadcd/sadcd-bootstrap.cpp
+++ b/src/sadcd/sadcd-bootstrap.cpp
@@ -14,6 +14,7 @@
 #include <fstream>
 #include <string>
 #include <string_view>
+ #include <unistd.h>  //execve(2)
 
 // external includes
 #include <boost/program_options.hpp>
@@ -97,16 +98,22 @@ display_help(po::options_description const& options)
 }
 
 void
-bootstrap(po::variables_map const& config)
+start_server(po::variables_map const& config)
 {
-	std::cerr << "Error: bootstrap() unimplemented\n";
-	std::exit(1);
-}
+	char port_str[2];
+	if (config.count("port"))
+	{
+		memcpy(port_str, &(config["port"].as<std::uint16_t>()), 2);
+	}else{
+		std::cerr << "Error: No port number was set\n";
+	}
+	const char *chunk_daemon = "sadcd";
+	char *const chunk_argv[] = {port_str, NULL};
 
-void
-start_server()
-{
-	std::cerr << "Error: start_server() unimplemented\n";
+	errno = 0;
+	execve(chunk_daemon, chunk_argv, NULL);
+	std::cerr << "Error: could not start chunk server: " << 
+		std::strerror(errno) << std::endl;
 	std::exit(1);
 }
 
@@ -135,8 +142,7 @@ main(int argc, char const** argv)
 	parse_config_file(variables, options);
 
 	// perform bootstrapping and start sadcd
-	bootstrap(variables);
-	start_server();
+	start_server(variables);
 
 	// start_server must not return since it is supposed to
 	// replace the program image with execve(2)
--- a/src/sadcd/sadcd-bootstrap.cpp
+++ b/src/sadcd/sadcd-bootstrap.cpp
@@ -101,8 +101,11 @@ display_help(po::options_description const& options)
 void
 start_server(po::variables_map const& config)
 {	
+	char sadcd_str[] = "sadcd";
+	char port_arg_str[] = "port";
+
 	//Verify a port number was given
-	if (!config.count("port"))
+	if (!config.count(port_arg_str))
 	{
 		std::cerr << "Error: No port number was set\n";
 		std::exit(1);
@@ -114,7 +117,7 @@ start_server(po::variables_map const& config)
 	strncpy(port_str, port.c_str(), sizeof(char) * port.size());
 	
 	// Arguments needed for execve
-	std::array<char *, 4> chunk_argv = {"sadcd", "port", port_str, NULL};
+	std::array<char *, 4> chunk_argv = {sadcd_str, port_arg_str, port_str, NULL};
 
 	execve(chunk_argv[0], &(chunk_argv[0]), NULL);
 
--- a/src/sadcd/sadcd-bootstrap.cpp
+++ b/src/sadcd/sadcd-bootstrap.cpp
@@ -100,23 +100,23 @@ display_help(po::options_description const& options)
 // Populates command-line args for sadcd and starts it
 void
 start_server(po::variables_map const& config)
-{
-	// Copy port from uint16 to C string
-	char port_str[2];
-	if (config.count("port"))
+{	
+	//Verify a port number was given
+	if (!config.count("port"))
 	{
-		memcpy(port_str, &(config["port"].as<std::uint16_t>()), 2);
-	}else{
 		std::cerr << "Error: No port number was set\n";
 		std::exit(1);
 	}
+
+	// Copy port from uint16 to C string
+	auto port = std::to_string(config["port"].as<std::uint16_t>());
+	char *port_str;
+	strncpy(port_str, port.c_str(), sizeof(char) * port.size());
 	
 	// Arguments needed for execve
-	const char *chunk_daemon = "sadcd";
-	char *const chunk_argv[] = {port_str, NULL};
+	std::array<char *, 4> chunk_argv = {"sadcd", "port", port_str, NULL};
 
-	errno = 0;
-	execve(chunk_daemon, chunk_argv, NULL);
+	execve(chunk_argv[0], &(chunk_argv[0]), NULL);
 
 	// Execve only returns if an error occured
 	std::cerr << "Error: could not start chunk server: " << 
--- a/src/sadcd/sadcd-bootstrap.cpp
+++ b/src/sadcd/sadcd-bootstrap.cpp
@@ -101,30 +101,23 @@ display_help(po::options_description const& options)
 void
 start_server(po::variables_map const& config)
 {	
-	char sadcd_str[] = "sadcd";
-	char port_arg_str[] = "port";
-
 	//Verify a port number was given
-	if (!config.count(port_arg_str))
+	if (!config.count("port"))
 	{
 		std::cerr << "Error: No port number was set\n";
 		std::exit(1);
 	}
-
-	// Copy port from uint16 to C string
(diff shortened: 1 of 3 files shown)


src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,30 @@ display_help(po::options_description const& options)
...
+start_server(po::variables_map const& config)
 {
-	std::cerr << "Error: bootstrap() unimplemented\n";
-	std::exit(1);
-}
+	// Copy port from uint16 to C string
+	char port_str[2];
+	if (config.count("port"))
+	{
+		memcpy(port_str, &(config["port"].as<std::uint16_t>()), 2);
```
damanmkumar: This doesn't do what we want it to do. `memcpy` copies the `std::uint16_t` byte-wise into a `char` array. What we want is a string representation of the port number, like `"6666"`, perhaps created using https://en.cppreference.com/w/cpp/string/basic_string/to_string.

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,30 @@ display_help(po::options_description const& options)
...
+	}else{
+		std::cerr << "Error: No port number was set\n";
+		std::exit(1);
+	}
+	
+	// Arguments needed for execve
+	const char *chunk_daemon = "sadcd";
+	char *const chunk_argv[] = {port_str, NULL};
```
damanmkumar: When `sadcd` starts running, its `argv` should look something like this:  
`{ "sadcd", "--port", "6666", nullptr }`

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,30 @@ display_help(po::options_description const& options)
...
+// Populates command-line args for sadcd and starts it
 void
-bootstrap(po::variables_map const& config)
+start_server(po::variables_map const& config)
 {
-	std::cerr << "Error: bootstrap() unimplemented\n";
-	std::exit(1);
-}
+	// Copy port from uint16 to C string
+	char port_str[2];
```
damanmkumar: Let's use `auto` instead of explicitly mentioning types wherever possible, and use `std::array` instead of naked, C-style arrays. Something like `auto chunk_argv = std::array<char const*, argc+1>`.

src/sadcd/sadcd-bootstrap.cpp
damanmkumar: As per convention, the first (0th) element in `argv` should be the name of the program.

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,30 @@ display_help(po::options_description const& options)
...
+	
+	// Arguments needed for execve
+	const char *chunk_daemon = "sadcd";
+	char *const chunk_argv[] = {port_str, NULL};
 
-void
-start_server()
-{
-	std::cerr << "Error: start_server() unimplemented\n";
+	errno = 0;
```
damanmkumar: The OS and the C standard library set `errno` when something goes wrong. So, we must use it as a read-only variable unless we write a C function that needs to indicate an error condition.

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,30 @@ display_help(po::options_description const& options)
...
 {
-	std::cerr << "Error: bootstrap() unimplemented\n";
-	std::exit(1);
-}
+	// Copy port from uint16 to C string
+	char port_str[2];
+	if (config.count("port"))
+	{
+		memcpy(port_str, &(config["port"].as<std::uint16_t>()), 2);
+	}else{
```
damanmkumar: This is really me nit-picking, but it'll probably be cleaner to check for a missing port number, and show an error if it *is* missing. We can do it with a single `if` block instead of `if` and `else`.

src/sadcd/sadcd.cpp
```diff
@@ -0,0 +1,8 @@
+#include <iostream>
```
damanmkumar, line 1: Let's add a comment at the top of the file to indicate what the code in this file does.

damanmkumar (changes requested):
Caught a couple of errors, and some general comments.

src/sadcd/sadcd-bootstrap.cpp
abmono93: Yes I see what you're saying but to_string will create an std::string, which I don't think execve will accept. But yes I need to figure out a way to get a C string representation of the port

src/sadcd/sadcd-bootstrap.cpp
abmono93: Yes I agree that is a better way to do it

src/sadcd/sadcd-bootstrap.cpp
damanmkumar: https://en.cppreference.com/w/cpp/string/basic_string/c_str is your friend :)

src/sadcd/sadcd-bootstrap.cpp
abmono93: Thanks, I just rewrote this whole section in a way that hopefully address this and your other comments. Lemme know what you think

src/sadcd/sadcd.cpp
abmono93: At the moment this file only needs to be here to verify if a sadcd-bootstrap call to execve() works. I will add some behavior to it for my next pull request and include a description

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,33 @@ display_help(po::options_description const& options)
...
+// Populates command-line args for sadcd and starts it
 void
-bootstrap(po::variables_map const& config)
-{
-	std::cerr << "Error: bootstrap() unimplemented\n";
-	std::exit(1);
-}
+start_server(po::variables_map const& config)
+{	
+	char sadcd_str[] = "sadcd";
```
damanmkumar: We don't need temporary variables to store a pointer to these string literals. They can be declared while initializing the `argv` array.

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,33 @@ display_help(po::options_description const& options)
...
+	char port_arg_str[] = "port";
```
damanmkumar: Should be `--port`

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,33 @@ display_help(po::options_description const& options)
...
+	if (!config.count(port_arg_str))
+	{
+		std::cerr << "Error: No port number was set\n";
+		std::exit(1);
+	}
+
+	// Copy port from uint16 to C string
+	auto port = std::to_string(config["port"].as<std::uint16_t>());
+	char *port_str;
+	strncpy(port_str, port.c_str(), sizeof(char) * port.size());
```
damanmkumar: 1. We don't need to copy the contents of the `port` into another buffer. All we need is a pointer to the buffer that `std::string` internally maintains, which is what `std::string::c_str()` gives us.
1. The `strncpy` here does not do what we want it to do, and would crash the program in the *best* case scenario.

src/sadcd/sadcd-bootstrap.cpp
abmono93: You're right it's not necessary but I changed it to be this way because declaring them in the array initializer causes a compilation warning

src/sadcd/sadcd-bootstrap.cpp
abmono93: I originally just had 
std::array<char` *, 4> chunk_argv = {sadcd_str, port_arg_str, port.c_str() , NULL};
but it doesn't allow me to compile it this way

abmono93: src/sadcd/sadcd-bootstrap.cpp:120:63: error: cannot initialize an array element of type 'char *'
      with an rvalue of type 'const std::__1::basic_string<char, std::__1::char_traits<char>,
      std::__1::allocator<char> >::value_type *' (aka 'const char *')
        std::array<char *, 4> chunk_argv = {sadcd_str, port_arg_str, port.c_str() , NULL};

src/sadcd/sadcd-bootstrap.cpp
abmono93: Thanks I'll change this

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,33 @@ display_help(po::options_description const& options)
...
+	
+	// Arguments needed for execve
+	std::array<char *, 4> chunk_argv = {sadcd_str, port_arg_str, port_str, NULL};
```
damanmkumar: String literals like `"sadcd"` and `"--port"` can be declared here as:
`auto chunk_argv = std::array<char const*, 4>{ "sadcd", "--port", port.c_str(), nullptr };`

Also `nullptr` is C++ equivalent to C's `NULL`.

src/sadcd/sadcd-bootstrap.cpp
```diff
@@ -96,17 +97,33 @@ display_help(po::options_description const& options)
...
+
+	execve(chunk_argv[0], &(chunk_argv[0]), NULL);
+
+	// Execve only returns if an error occured
+	std::cerr << "Error: could not start chunk server: " <<
```
damanmkumar: Wait, since `main` shows an error message if `start_server` returns, we don't need the error handling here. Or in `main`. Your call!

damanmkumar (changes requested):
A few more comments

src/sadcd/sadcd-bootstrap.cpp
damanmkumar: Yup, that's because `port.c_str()` returns a `char const*` since it doesn't expect us to overwrite the string it contains/owns (when we use `c_str()`). And if you think about it, `execve` doesn't ask for a mutable `char*` since it only reads the strings, and doesn't overwrite them. So, `chunk_argv` should be an array of pointers to immutable `char`s, aka `char const*`.

src/sadcd/sadcd-bootstrap.cpp
abmono93: We could do it this way but we'd get a bunch of warnings every time we run make:

Adams-MBP-3:sadfs adammonahan$ make chunk
g++ -c -std=c++17 -ggdb -O -Iinclude        src/sadcd/sadcd-bootstrap.cpp -o build/sadcd/sadcd-bootstrap.o
src/sadcd/sadcd-bootstrap.cpp:120:38: warning: ISO C++11 does not allow conversion from string literal to 'char *' [-Wwritable-strings]
        std::array<char *, 4> chunk_argv = {"sadcd", "--port", port_str , NULL};
                                                                     ^
src/sadcd/sadcd-bootstrap.cpp:120:47: warning: ISO C++11 does not allow conversion from string literal to 'char *' [-Wwritable-strings]
        std::array<char *, 4> chunk_argv = {"sadcd", "--port", port_str , NULL};
                                                                                    ^
2 warnings generated.

src/sadcd/sadcd-bootstrap.cpp
abmono93: Why would `strncpy`cause the program crash?

src/sadcd/sadcd-bootstrap.cpp
damanmkumar: See https://github.com/CS297P-SAD/sadfs/pull/19#discussion_r513878175

src/sadcd/sadcd-bootstrap.cpp
abmono93: Yes that's true. I found having the message here helpful while debugging because execve gave me hints as to what I did wrong with errno. Although I suppose it would still be available after it returns to main

src/sadcd/sadcd-bootstrap.cpp
damanmkumar: When we declare `port_str` as a `char*`, it is not initialized, and will contain some garbage data (say `0x326b123e`). Then, `strncpy` will go and write the data contained in the string `port` at the memory location pointed to by `port_str`. If the program doesn't have access to `0x326b123e`, it will crash. If it does have write access to that memory location, however, it will silently overwrite some data elsewhere in the program.  

`strncpy` expects the address of a buffer. `port_str` should really be a `char` array if we were to use this.

damanmkumar (approved):
Let's get this moving, haha

PR #21, opened 2020-11-03
Chunk start conns

### LINKED ISSUE #11: Chunk server daemon should be able to establish new connections
The Chunk server daemon should be able to establish new connections to the Master server and other Chunk servers

abmono93:
Doesn't start conns with other chunk servers as this shouldn't happen on startup, and a new class for outgoing messages should be created first.

Closes #11

### DIFF
--- a/src/comm/inet.cpp
+++ b/src/comm/inet.cpp
@@ -15,25 +15,7 @@
 namespace sadfs { namespace inet {
 
 ip_addr::
-ip_addr()
-{
-	addr_ = 0;
-}
-
-ip_addr::
 ip_addr(char const* ip)
-{
-	set(ip);
-}
-
-std::uint32_t ip_addr::
-value() const noexcept
-{
-	return addr_;
-}
-
-void ip_addr::
-set(char const* ip)
 {
 	auto tmp = in_addr{};
 	if (inet_aton(ip, &tmp) == 0)
@@ -43,6 +25,12 @@ set(char const* ip)
 		throw std::invalid_argument(err.str());
 	}
 	addr_ = tmp.s_addr;
+}
+
+std::uint32_t ip_addr::
+value() const noexcept
+{
+	return addr_;
 }
 
 namespace {
--- a/src/comm/inet.cpp
+++ b/src/comm/inet.cpp
@@ -15,7 +15,25 @@
 namespace sadfs { namespace inet {
 
 ip_addr::
+ip_addr()
+{
+	addr_ = 0;
+}
+
+ip_addr::
 ip_addr(char const* ip)
+{
+	set(ip);
+}
+
+std::uint32_t ip_addr::
+value() const noexcept
+{
+	return addr_;
+}
+
+void ip_addr::
+set(char const* ip)
 {
 	auto tmp = in_addr{};
 	if (inet_aton(ip, &tmp) == 0)
@@ -25,12 +43,6 @@ ip_addr(char const* ip)
 		throw std::invalid_argument(err.str());
 	}
 	addr_ = tmp.s_addr;
-}
-
-std::uint32_t ip_addr::
-value() const noexcept
-{
-	return addr_;
 }
 
 namespace {
--- a/src/sadcd/sadcd.cpp
+++ b/src/sadcd/sadcd.cpp
@@ -7,70 +7,72 @@
 
 using namespace sadfs;
 
-void
-parse_args(int argc, char *argv[])
+//struct to hold info needed to set up a chunk server
+struct chunk_server_settings
 {
-	for (auto i = 1; i < argc - 1; i += 2)
+	std::uint16_t port;
+	void from_command_line(int, char**);
+};
+
+//Sets struct variables basd on command line arguments passed
+void chunk_server_settings::
+from_command_line(int argc, char *argv[])
+{
+	for (auto i = 1; i < argc - 1; i ++)
 	{
 		auto arg = std::string{argv[i]};
-		auto val = std::string{argv[i + 1]};
-		//Set attribute to val basd on arg
-	}
-}
-
-//Temporary
-std::uint16_t
-get_port(int argc, char *argv[])
-{
-	auto port_flag = std::string{"--port"};
-
-	for (auto i = 1; i < argc - 1; i += 2)
-	{
-		auto arg = std::string{argv[i]};
-		auto val = std::string{argv[i + 1]};
-		if (port_flag.compare(arg) == 0)
+		if (arg.compare("--port") == 0)
 		{
-			return static_cast<std::uint16_t>(std::stoi(val));
+			i++;
+			this->port = static_cast<std::uint16_t>(std::stoi(std::string{argv[i]}));
 		}
 	}
-
-	return 0;
 }
 
-//struct
+//Reads the message from a socket that just received some data
+std::string
+process_message(sadfs::socket* sock)
+{
+	auto buf = std::array<char, 512>{};
+	auto len = 0;
+	auto result = std::string{};
+	while ((len = ::read(sock->descriptor(), buf.data(), buf.size() - 1)))
+	{
+		if (len == -1)
+		{
+			std::cerr << "read error\n";
+			std::cerr << std::strerror(errno) << std::endl;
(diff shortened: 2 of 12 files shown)


src/comm/inet.cpp
```diff
@@ -14,8 +14,26 @@
 
 namespace sadfs { namespace inet {
 
+ip_addr::
```
damanmkumar: The reason behind making `ip_addr` a `class`, and not a `struct`, was to validate the ip address that is passed to it while constructing the object. What a default constructor that assigns `0` to `addr_` does is implicitly construct `ip_addr` with `"0.0.0.0"`.

`ip_addr` doesn't have a default constructor by design. Due to the nature of ip addresses, and the fact that the ip addresses of the Master, and other Chunk servers do not change, it doesn't make sense to construct an `ip_addr` with one value and "set" it's value to something else.

Lastly, they are pretty cheap to construct and have implicitly defined copy constructors. So, 
`object.ip = {"10.0.0.1"};` is perfectly valid.

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,92 @@
+/* This creates the binary for the chunk server daemon, which should only be
+ * started by the sadcd-bootstrap process. It opens up a listener on an ip and
+ * port specified by the command line to handle incoming requests.
+ */
+
+#include <sadfs/comm/inet.hpp>
+
 #include <iostream>
+#include <array>
```
damanmkumar, line 10: Let's sort `include`s in alphabetical order?

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,92 @@
...
+struct chunk_server_settings
+{
+	std::uint16_t port;
+	sadfs::inet::ip_addr ip;
+	void from_command_line(int, char**);
+};
+
+// sets struct variables based on command line arguments passed
+void chunk_server_settings::
+from_command_line(int argc, char *argv[])
```
damanmkumar: How about a function that returns an object of type `chunk_server_settings` instead of making this a function of the `struct`?

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,92 @@
...
+#include <cstdint>  // std::uint16_t
+
+using namespace sadfs;
+
+// struct to hold info needed to set up a chunk server
+struct chunk_server_settings
+{
+	std::uint16_t port;
+	sadfs::inet::ip_addr ip;
+	void from_command_line(int, char**);
```
damanmkumar: The function will go if we make `from_command_line` a function that returns a `chunk_server_settings`. Also, if we only use the ip and port number to create a `listener` and forget about it, we can throw this data structure in an unnamed namespace, and just name it `settings` or `config`.

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,92 @@
...
+void chunk_server_settings::
+from_command_line(int argc, char *argv[])
+{
+	for (auto i = 1; i < argc - 1; i ++)
+	{
+		auto arg = std::string{argv[i]};
+		if (arg.compare("--port") == 0)
+		{
+			i++;
+			this->port = static_cast<std::uint16_t>(std::stoi(std::string{argv[i]}));
```
damanmkumar: We either need to validate that the port number is actually within the limits of an unsigned 16-bit number, or log the port number (which would be a different user story). If we want to take the second route, we should document it with a comment, since the `static_cast` will essentially do `port % 65535`. I feel it is an error and we must report an error and exit in that case. 

Since this will be an issue only if someone runs `sadcd` by itself and not through `sadcd-bootstrap`, we totally can just create an issue on GitHub and fix it if we have time towards the end of the quarter. In which case, we should totally add a comment here.

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,92 @@
...
+		{
+			i++;
+			this->ip.set(argv[i]);
+		}
+	}
+}
+
+// reads the message from a socket that just received some data
+std::string
+process_message(sadfs::socket* sock)
```
damanmkumar: Passing `socket` by reference will prevent an issue with possibly sending in a `nullptr`.

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,92 @@
...
+
+	// set up a listener for incoming messages
+	auto listener = inet::listener{chunk_server.ip, chunk_server.port};
+
+	while (true)
+	{
+		auto sock = listener.accept();
+		auto result = process_message(&sock);
+
+		std::cout << result << std::endl;
```
damanmkumar: `std::endl` flushes the buffer in addition to adding a newline character, and is totally unnecessary. `'\n'` or (preferably, in this case) `"\n"` is more appropriate.

damanmkumar (changes requested):
Just one major design change, and minor gripes here and there.

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,92 @@
...
+ */
+
+#include <sadfs/comm/inet.hpp>
+
 #include <iostream>
+#include <array>
+#include <unistd.h> // read/write
+#include <cstdint>  // std::uint16_t
+
+using namespace sadfs;
```
damanmkumar: It's good practice to not pollute the global namespace with other namespaces. This pulls everything from the sadfs namespace into the global namespace. We can place all these functions (except main) within `namespace sadfs {}`. All items from the sadfs namespace will be available within this block as well.

src/comm/inet.cpp
abmono93: Thanks, I wasn't sure how to do this without a default constructor but I think I see what you're saying now. I'll move some things around and see if that's what you had in mind.

src/sadcd/sadcd.cpp
abmono93: Didn't think of that, thanks

src/sadcd/sadcd.cpp
abmono93: Actually I'm just going to get rid of this in that case. I only use need sadfs objects once so far so will just expilicity use sadfs::inet...

src/sadcd/sadcd.cpp
abmono93: Actually I just had a thought, another nice way to do this could be to handle the error checking the same way we do with the ip address, i.e. have the sadfs::inet::port_no take in a string or regular int and do the checking/conversion itself. That seems to work nicely with my testing of ip ports so far

damanmkumar: `sadfs::inet::port_no` is just an alias for `std::uint16_t` unfortunately. We could totally turn it into a class that performs this bounds checking for us.

src/sadcd/sadcd.cpp
damanmkumar: Yup, that works as well. But if you see it getting out of hand as we add more functionality to `sadcd`, this might be a better approach.

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,102 @@
...
 #include <iostream>
+#include <unistd.h> // read/write
+
+namespace
+{
+
+// struct to hold info needed to set up a chunk server
+struct settings
+{
+	std::uint16_t port;
```
damanmkumar: We should use the more appropriate type `sadfs::inet::port_no` here

src/sadcd/sadcd.cpp
```diff
@@ -1,8 +1,102 @@
...
 
 int 
-main(int, char **)
+main(int argc, char *argv[])
 {
-	std::cout << "Server started" << std::endl;
-	//TODO: open ports
+
+	// populate chunk_server from command line args
+	auto chunk_server = read_command_line(argc, argv);
```
damanmkumar: Maybe `settings` or `server_settings` is a more descriptive name here. Also, we might need to use the name `chunk_server` if we decide to introduce a class to represent a Chunk server.

damanmkumar (changes requested):
Just two very minor comments. Everything else looks good.

damanmkumar (approved):

PR #23, opened 2020-11-03
moved build config to CMake from manually written Makefile

damanmkumar:
* moved bootstrap files (except common `util.cpp`) to corresponding daemon directory
* also fixes `sadcd` build error on Linux

**NOTE:** be sure to squash and merge


include/sadfs/comm/inet.hpp
```diff
@@ -2,7 +2,7 @@
 #define SADFS_COMM_INET_HPP
 
 // sadfs-specific includes
-#include <sadfs/comm/socket.hpp>
+#include "socket.hpp"
```
abmono93, line 5: Is there a specific reason we need to import it this way instead?

abmono93 (approved):

PR #24, opened 2020-11-07
Issue10 master listen

### LINKED ISSUE #10: Master daemon should listen for connections
The Master server daemon should listen on a configured port number for connections from Clients and Chunk servers

abmono93:
Closes #10


src/sadmd/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+		{
+			i++;
+			ip = argv[i];
+		}
+	}
+
+	// TODO: verify that port fits in 16 bits in edge case that this program is
+	// started by some method other than sadcd-bootstrap
+	// should also double check that an ip and port number were in fact passed
+	return settings{static_cast<std::uint16_t>(port), std::string(ip).c_str()};
```
damanmkumar: The `std::string(ip).c_str()` here will lead to undefined behaviour since the string itself is destroyed after construction of the `settings` object. So, `settings::ip` will be a dangling reference. `settings{..., ip}` should work, right? I see no reason to construct a `std::string` in this case.

src/sadmd/sadmd.cpp
```diff
@@ -1,16 +1,56 @@
...
+		// perform some action based on result
+	}
+}
+
+std::string sadmd::
+process_message(sadfs::socket& sock)
+{
+	auto buf = std::array<char, 512>{};
+	auto len = 0;
+	auto result = std::string{};
```
damanmkumar, line 34: Just a comment - it is irrelevant for this use-case (since this function will go away once we implement `sadmd`), but in any other situation, it would help to have a call to `result.reserve()` (https://en.cppreference.com/w/cpp/string/basic_string/reserve). That would avoid repeated resizing of the string on line 50.

damanmkumar (changes requested):
One review comment + one note.

src/sadmd/main.cpp
abmono93: Yea you're right that does work. I swear I tried dong this so many different ways.... don't know why that wasn't one of them

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef SADFS_SADMD_SADMD_HPP
+#define SADFS_SADMD_SADMD_HPP
+#include <sadfs/comm/inet.hpp>
+
+#include <string>
+
+class sadmd
+{
+public:
+    sadfs::inet::port_no port;
```
damanmkumar: There's no reason to expose the port number and IP address as public members here, right?

damanmkumar (changes requested):
Missed this, sorry!

include/sadfs/sadmd/sadmd.hpp
abmono93: It allows you to construct a sadmd object with: auto master = sadmd{ip, port}
Otherwise you have to write a constructor which I was unable to do because of the way ip_addr objects are initialized. I'm open to doing it with a constructor if you can figure out how

damanmkumar: I see! Something like this should work:
```
class {
public:
    sadmd(char const*, port_no);
private:
    ip_addr ip_;
    port_no port_;
};

sadmd::
sadmd(char const* ip, port_no port) : ip_(ip), port_(port)
{
   // do nothing
}
```

If you want `sadmd` to accept an `ip_addr` directly, this should work:
```
auto master = sadmd{{"127.0.0.1"}, port};
```

PR #25, opened 2020-11-09
Created sadcd class and moved chunk daemon into main.cpp file

abmono93:
To be consistent with changes made to chunkmd for #10 
creates new class chunkcd to be used by main process
(branch name was a mistake- it should be called chunkcd-class)


damanmkumar (approved):
fixed a minor typo

Issue #13, opened 2020-10-23
Introduce class to establish new connections


damanmkumar:
Introduce a class to easily establish a new connection to another entity

damanmkumar:
closed with #26

PR #26, opened 2020-11-10
implemented outbound connection establishment

damanmkumar:
* `service` class abstracts a service running on a (remote) host
* `service::connect` lets us establish a connection to the service
* some minor refactoring

implements #13 

**NOTE:** squash and merge

### DIFF
--- a/include/sadfs/comm/inet.hpp
+++ b/include/sadfs/comm/inet.hpp
@@ -5,7 +5,6 @@
 #include "socket.hpp"
 
 // standard includes
-#include <cstddef> // std::size_t
 #include <cstdint> // std::uint32_t, std::uint16_t
 
 namespace sadfs { namespace inet {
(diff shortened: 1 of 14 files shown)


include/sadfs/comm/inet.hpp
```diff
@@ -42,14 +69,13 @@ class listener
 	// accepts a new connection and returns a socket
 	socket accept() const;
 private:
-	ip_addr ip_;
-	port_no port_;
-	socket  socket_;
+	service const service_;
+	socket        socket_;
```
abmono93, line 73: What is socket_ for?

abmono93 (approved):

include/sadfs/comm/inet.hpp
damanmkumar: `listener` needs to open a socket and bind it to the IP address and port number that is specified, using the `bind(2)` system call. The socket is then marked for listening, which will let us actually accept connections.

PR #29, opened 2020-11-13
Create sadmd data structures, database and load database contents

### LINKED ISSUE #28: Master should create persistent and in-memory storage for filesystem metadata on startup
- Persistent storage of all filnames on the system
- In-memory storage for active chunk servers
- In-memory storage for which chunks are on which chunk servers

abmono93:
The exact info sadmd keeps about files and chunk servers is based on what is currently in ideas.md but will likely change and it should be straigtforward to do so.

If you manuallly add some rows to the database this will print the filenames and chunkid lists as there is no other way to verify that this works but these print statements should be removed once there is some more functionality associated with the files_ object

Closes #28

### DIFF
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -14,7 +14,8 @@ namespace sadfs {
 using chunkid = uint64_t;
 
 // all the information needed about a chunk server
-struct chunk_server_info{
+struct chunk_server_info
+{
 	inet::service service;
 	uint64_t max_chunks;
 	uint64_t chunk_count;
@@ -22,7 +23,8 @@ struct chunk_server_info{
 };
 
 // all the information needed about a file
-struct file_info{
+struct file_info
+{
 	int ttl;
 	std::vector<chunkid> chunkids;
 };
@@ -34,11 +36,14 @@ public:
 
 	// starts server
 	void start();
+
+private:
 	// creates (the metadata for) a new file
 	void create_file(std::string const&);
-private:
+
 	// loads file metadata from disk
 	void load_files();
+
 	// runs an sql statement on system_files_
 	void db_command(std::string const&) const noexcept;
 
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -25,8 +25,6 @@ struct chunk_server_info{
 struct file_info{
 	int ttl;
 	std::vector<chunkid> chunkids;
-	// decides whether file is currently locked for writing
-	bool locked();
 };
 
 class sadmd
@@ -39,24 +37,20 @@ public:
 	// creates (the metadata for) a new file
 	void create_file(std::string const&);
 private:
-	// reads the message from a socket that just received some data
-	std::string process_message(sadfs::socket const&);
 	// loads file metadata from disk
 	void load_files();
 	// runs an sql statement on system_files_
 	void db_command(std::string const&) const noexcept;
 
-	const inet::service service_;
+	inet::service const service_;
 	// in memory representation of each file
 	std::unordered_map<std::string, file_info> files_;
-	// list of active chunk servers
-	std::unordered_map<uint64_t, chunk_server_info> chunk_servers_;
+	// metadata for each chunk server
+	std::unordered_map<uint64_t, chunk_server_info> chunk_server_metadata_;
 	// map from chunkid to list of chunk servers
-	std::unordered_map<chunkid, std::vector<chunk_server_info*> > chunkids_;
+	std::unordered_map<chunkid, std::vector<chunk_server_info*> > chunk_locations_;
 	// persistent/on disk copy of files_
 	sqlite3* const files_db_;
-	sqlite3* open_db();
-
 };
 
 } // sadfs namespace
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -3,9 +3,29 @@
 #include <sadfs/comm/inet.hpp>
 #include <sadfs/comm/socket.hpp>
 
+#include <ctime>
 #include <string>
+#include <sqlite3.h>
+#include <unordered_map>
+#include <vector>
 
 namespace sadfs {
+
+// all the information needed about a chunk server
+struct chunk_server_info{
+	inet::ip_addr ip;
+	inet::port_no port;
+	int total_space;
+	int used_space;
+	time_t ttl;
+};
+
+// all the information needed about a file
+struct file_info{
+	bool locked;
+	time_t ttl;
+	std::vector<int> chunkids;
+};
 
 class sadmd
 {
@@ -14,11 +34,26 @@ public:
 
 	// starts server
 	void start();
+	// creates (the metadata for) a new file
+	void create_file(std::string);
 private:
 	// reads the message from a socket that just received some data
 	std::string process_message(sadfs::socket const&);
+	// loads file metadata from disk
+	void load_files();
+	// runs an sql statement on system_files_
+	void db_command(std::string);
 
 	inet::service const service_;
+	// in memory representation of each file
+	std::unordered_map<std::string, file_info> files_;
+	// list of active chunk servers
+	std::vector<chunk_server_info> active_chunk_servers_;
+	// map from chunkid to list of chunk servers
+	std::unordered_map<int, std::vector<chunk_server_info*> > chunkids_;
+	// persistent/on disk copy of files_
+	sqlite3* files_db_;
+
 };
 
 } // sadfs namespace
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -11,13 +11,13 @@
 
 namespace sadfs {
 
-using chunkid = unsigned int;
+using chunkid = uint64_t;
 
 // all the information needed about a chunk server
 struct chunk_server_info{
 	inet::service service;
-	unsigned int total_space;
-	unsigned int used_space;
+	uint64_t total_space;
+	uint64_t used_space;
 	int ttl;
 };
 
@@ -50,7 +50,7 @@ private:
 	// in memory representation of each file
 	std::unordered_map<std::string, file_info> files_;
 	// list of active chunk servers
-	std::unordered_map<unsigned int, chunk_server_info> chunk_servers_;
+	std::unordered_map<uint64_t, chunk_server_info> chunk_servers_;
 	// map from chunkid to list of chunk servers
 	std::unordered_map<chunkid, std::vector<chunk_server_info*> > chunkids_;
 	// persistent/on disk copy of files_
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -16,8 +16,8 @@ using chunkid = uint64_t;
 // all the information needed about a chunk server
 struct chunk_server_info{
 	inet::service service;
-	uint64_t total_space;
-	uint64_t used_space;
+	uint64_t max_chunks;
+	uint64_t number_stored_chunks;
 	int ttl;
 };
 
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -17,7 +17,7 @@ using chunkid = uint64_t;
 struct chunk_server_info{
 	inet::service service;
 	uint64_t max_chunks;
-	uint64_t number_stored_chunks;
+	uint64_t chunk_count;
 	int ttl;
 };
 
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -50,7 +50,7 @@ private:
 	// in memory representation of each file
 	std::unordered_map<std::string, file_info> files_;
 	// list of active chunk servers
-	std::vector<chunk_server_info> chunk_servers_;
+	std::unordered_map<unsigned int, chunk_server_info> chunk_servers_;
 	// map from chunkid to list of chunk servers
 	std::unordered_map<chunkid, std::vector<chunk_server_info*> > chunkids_;
 	// persistent/on disk copy of files_
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -3,7 +3,7 @@
 #include <sadfs/comm/inet.hpp>
 #include <sadfs/comm/socket.hpp>
 
-#include <ctime>
+#include <chrono>
 #include <string>
 #include <sqlite3.h>
 #include <unordered_map>
@@ -11,20 +11,22 @@
 
 namespace sadfs {
 
+using chunkid = unsigned int;
+
 // all the information needed about a chunk server
 struct chunk_server_info{
-	inet::ip_addr ip;
-	inet::port_no port;
-	int total_space;
-	int used_space;
-	time_t ttl;
+	inet::service service;
+	unsigned int total_space;
+	unsigned int used_space;
+	int ttl;
 };
 
 // all the information needed about a file
 struct file_info{
-	bool locked;
-	time_t ttl;
-	std::vector<int> chunkids;
+	int ttl;
+	std::vector<chunkid> chunkids;
+	// decides whether file is currently locked for writing
+	bool locked();
 };
 
 class sadmd
@@ -35,24 +37,25 @@ public:
 	// starts server
 	void start();
 	// creates (the metadata for) a new file
-	void create_file(std::string);
+	void create_file(std::string const&);
 private:
 	// reads the message from a socket that just received some data
 	std::string process_message(sadfs::socket const&);
 	// loads file metadata from disk
 	void load_files();
 	// runs an sql statement on system_files_
-	void db_command(std::string);
+	void db_command(std::string const&) const noexcept;
 
-	inet::service const service_;
+	const inet::service service_;
 	// in memory representation of each file
 	std::unordered_map<std::string, file_info> files_;
 	// list of active chunk servers
-	std::vector<chunk_server_info> active_chunk_servers_;
+	std::vector<chunk_server_info> chunk_servers_;
 	// map from chunkid to list of chunk servers
-	std::unordered_map<int, std::vector<chunk_server_info*> > chunkids_;
+	std::unordered_map<chunkid, std::vector<chunk_server_info*> > chunkids_;
 	// persistent/on disk copy of files_
-	sqlite3* files_db_;
+	sqlite3* const files_db_;
+	sqlite3* open_db();
 
 };
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -30,6 +30,8 @@ add_executable (sadmd
 	sadmd/main.cpp
 	sadmd/sadmd.cpp
 	${COMMON_SOURCE_FILES})
+target_link_libraries(sadmd
+	sqlite3)
 add_executable (sadcd
 	sadcd/main.cpp
 	sadcd/sadcd.cpp
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -30,8 +30,6 @@ add_executable (sadmd
 	sadmd/main.cpp
 	sadmd/sadmd.cpp
 	${COMMON_SOURCE_FILES})
-target_link_libraries(sadmd
-	sqlite3)
 add_executable (sadcd
 	sadcd/main.cpp
 	sadcd/sadcd.cpp
@@ -39,6 +37,9 @@ add_executable (sadcd
 # add_executable (sadfsd
 #	sadfsd/sadfsd.cpp
 #	${COMMON_SOURCE_FILES})
+
+target_link_libraries(sadmd
+	sqlite3)
 
 # first, find the boost_program_options library. this sets
 # the Boost_INCLUDE_DIRS and Boost_LIBRARIES variables
--- a/src/sadmd/sadmd.cpp
+++ b/src/sadmd/sadmd.cpp
@@ -112,14 +112,13 @@ void sadmd::
 load_files()
 {
 	std::cout << "Files loaded:\n";
-	auto res = 0;
 	sqlite3_stmt* stmt;
 	if (sqlite3_prepare_v2(files_db_, "SELECT * FROM files;", -1, &stmt, NULL) != 0)
 	{
 		std::cerr << sqlite3_errmsg(files_db_) << '\n';
 	}
 
-	while ((res = sqlite3_step(stmt)) == SQLITE_ROW)
+	while (sqlite3_step(stmt) == SQLITE_ROW)
 	{
 		auto filename = std::string{reinterpret_cast<const char*>(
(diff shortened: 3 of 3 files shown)


include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
 #include <sadfs/comm/inet.hpp>
 #include <sadfs/comm/socket.hpp>
 
+#include <ctime>
```
damanmkumar: Let's use the more modern `std::chrono` here?

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
...
 #include <string>
+#include <sqlite3.h>
+#include <unordered_map>
+#include <vector>
 
 namespace sadfs {
 
+// all the information needed about a chunk server
+struct chunk_server_info{
+	inet::ip_addr ip;
```
damanmkumar: `inet::service` is a better fit here since the chunk server really provides a service, and it is intended to abstract IP + port number into a logical entity.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
...
+	inet::port_no port;
+	int total_space;
```
damanmkumar: `int` is small and signed. It might make sense to use a 4-byte integer to save memory, while counting number of chunks currently stored, and max number of chunks that the server can accommodate. `unsigned` or `std::size_t` will be better.

include/sadfs/sadmd/sadmd.hpp
abmono93: Yea good idea

include/sadfs/sadmd/sadmd.hpp
abmono93: Well yes but the main use of this will be to send this info to a client so they know where to read or write from. Would it make sense to send an `inet::service` object over the network?

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
...
+	int used_space;
+	time_t ttl;
```
damanmkumar: Since we'll be using the `std::chrono` library, we can use a regular integer to store the TTL.
There are ways to encode the current time into a regular integer type. While setting the TTL,
we can store the time, say ten minutes from now, into such an integer. That way, we can determine when the `{chunk_server|file}_info` has expired.
[`std::chrono::steady_clock`](https://en.cppreference.com/w/cpp/chrono/steady_clock) is recommended when dealing with the time duration between events.

Here's something I quickly put together. We will probably need better names.
https://godbolt.org/z/WhbcYf

include/sadfs/sadmd/sadmd.hpp
abmono93: I wasn't sure about this since I didn't know if this should mean number of bytes, kb or what so I went with the most straigtforward thing. But now that you say it number of chunks is a nice suggestion. And also yes it can be unsigned but not sure we would save much space using `std::size_t` as this will be one per chunk server

include/sadfs/sadmd/sadmd.hpp
damanmkumar: We will be sending the IP and port number in a control message. We can use `ip()` and `port()` to retrieve them anyway. Also, using `inet::service` has the added advantage of making sure there is no possibility of IPs and port numbers getting mixed up between different objects.

damanmkumar: Also, we should be sending the ip as a string, and the port number as a regular 16-bit number over the network. So, `ip_addr` should either store the string passed to it, or convert the network byte representation into a `std::string` on demand. Same with the port number.

include/sadfs/sadmd/sadmd.hpp
damanmkumar: We will save space if we use `unsigned` since it is 4 bytes on most platforms. `std::size_t` is 8 bytes on 64-bit platforms, which are the most common.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
...
+};
+
+// all the information needed about a file
+struct file_info{
+	bool locked;
+	time_t ttl;
```
damanmkumar: Is the TTL here for the lock held by a client? If so, we can get rid of the `bool`. If the TTL is "valid," the lock is held. Otherwise, it is free. What do you think?

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
...
+	std::vector<int> chunkids;
```
damanmkumar: Chunk IDs should be globally unique numbers. They should definitely be unsigned integers.

src/CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ add_executable (sadmd
 	sadmd/main.cpp
 	sadmd/sadmd.cpp
 	${COMMON_SOURCE_FILES})
+target_link_libraries(sadmd
```
damanmkumar: Let's move this further down the file so that things are organized

src/sadmd/sadmd.cpp
```diff
@@ -63,4 +90,38 @@ process_message(sadfs::socket const& sock)
 	return result;
 }
 
+void sadmd::
+db_command(std::string stmt)
```
damanmkumar: Since we're not modifying the string, and since we may call this function by passing a string constructed at runtime, let's pass the parameter `stmt` by `const&` to avoid making a copy. The `const` will help the compiler perform any possible optimizations.

src/sadmd/sadmd.cpp
```diff
@@ -63,4 +90,38 @@ process_message(sadfs::socket const& sock)
...
+{
+	auto chk = sqlite3_exec(files_db_, stmt.c_str(), NULL, NULL, NULL);
```
damanmkumar: Can we do away with `chk` here? How about something like `if (sqlite_exec(...) != 0)` ?

src/sadmd/sadmd.cpp
```diff
@@ -9,14 +9,29 @@
...
+#include <sqlite3.h>
 #include <unistd.h> // read/write
 
 namespace sadfs {
 
 sadmd::
 sadmd(char const* ip, int port) : service_(ip, port)
 {
-	// do nothing
+	sqlite3* db;
```
damanmkumar, line 20: ~Is this temporary variable necessary?~
If `files_db_` is made `const`, a helper function will need a temporary `sqlite3*`.

src/sadmd/sadmd.cpp
```diff
@@ -9,14 +9,29 @@
...
+
+	// open the sadmd database
+	auto chk = sqlite3_open("sadmd.db", &db);
+	if (chk)
+	{
+		std::cerr << "Error: Couldn't open database: ";
```
damanmkumar, line 25: This error condition needs to be handled. Now, the program will continue to execute and run `db_command` even when `sqlite3_open` fails.

src/sadmd/sadmd.cpp
```diff
@@ -31,9 +46,21 @@ start()
 
 		std::cout << result << "\n";
 		// perform some action based on result
+		// create_file(result);
 	}
 }
 
+void sadmd::
+create_file(std::string filename)
```
damanmkumar: passing the filename parameter by value unnecessarily makes a copy of the `std::string` here. `std::string const&` will avoid that

src/sadmd/sadmd.cpp
```diff
@@ -31,9 +46,21 @@ start()
...
+{
+	auto info = file_info{false, time(0), };
+	if (!files_.count(filename))
```
damanmkumar, line 91: Should this method handle the edge case where it is invoked with a file that already exists? Is it an error when such a thing happens?

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
...
 
 	inet::service const service_;
+	// in memory representation of each file
+	std::unordered_map<std::string, file_info> files_;
+	// list of active chunk servers
+	std::vector<chunk_server_info> active_chunk_servers_;
+	// map from chunkid to list of chunk servers
+	std::unordered_map<int, std::vector<chunk_server_info*> > chunkids_;
+	// persistent/on disk copy of files_
+	sqlite3* files_db_;
```
damanmkumar: If we never reassign `files_db_` to another database throughout the lifetime of `sadmd`, `files_db_` should be `const`. Introducing a method that performs the necessary setup and error handling like the `parse_*` methods in inet.cpp will help us initialize this member variable.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
...
 	void start();
+	// creates (the metadata for) a new file
+	void create_file(std::string);
 private:
 	// reads the message from a socket that just received some data
 	std::string process_message(sadfs::socket const&);
+	// loads file metadata from disk
+	void load_files();
+	// runs an sql statement on system_files_
+	void db_command(std::string);
```
damanmkumar: Since `db_command` does not modify any of the member variables in this class, it should be marked `const`. And all functions that do not throw exceptions should be marked `noexcept`.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,57 @@
...
 
 	inet::service const service_;
+	// in memory representation of each file
+	std::unordered_map<std::string, file_info> files_;
+	// list of active chunk servers
+	std::vector<chunk_server_info> active_chunk_servers_;
```
damanmkumar: When we receive updates/heartbeats from chunk servers, we will need to update the TTL in these info objects. If we use a `std::vector`, we'll need to linearly scan it to find the info object, which is an `O(n)` operation. Since this will happen continuously, it will not scale. We'll be better off using a faster data structure.

damanmkumar (changes requested):
I left a bunch of comments, most of which are minor

include/sadfs/sadmd/sadmd.hpp
abmono93: At one point I had this as a function `bool locked()` which would decide based on ttl. But then when thinking about how create_file() would work I realized files should be unlocked by default. So in order for that to work we'd need to decide on some convention for what the appropriate ttl is for a file that has never been locked. A straightforward way to do it which I only now thought of would be to set the ttl to one second  in the past, then locked() will return false iff ttl is before the current time. How's that sound?

src/sadmd/sadmd.cpp
abmono93: `std::string const&` or `const std::string&` ?

src/sadmd/sadmd.cpp
abmono93: Right. Do you think just exiting the program works?

src/sadmd/sadmd.cpp
abmono93: I was thinking this should just behave the way it does on a terminal. e.g. if I run  `touch tmp` twice the second time has no effect. And if I run `touch existingfile` the contents of existingfile remain unchanged. So I would argue it should not be not an error

include/sadfs/sadmd/sadmd.hpp
abmono93: We need a variable of type `sqlite3*` to pass to functions like `sqlite3_open()` and `sqlite3_exec`

include/sadfs/sadmd/sadmd.hpp
damanmkumar: A function is an awesome idea. The specification for `std::chrono::steady_clock` says that it always monotonically increases. A TTL of `0` would implicitly mean that the file has been free since the beginning of time.

src/sadmd/sadmd.cpp
damanmkumar: I've used the first convention throughout the code base so far. So, let's go with that!

src/sadmd/sadmd.cpp
damanmkumar: It is a fatal error if the server cannot read the database. So, exiting sounds like the right approach.

include/sadfs/sadmd/sadmd.hpp
abmono93: I think that depends on what these heartbeat messages will have in them. Would we just hash the ip? Or maybe there is a better way to do that, e.g. if retrieving the ip address of the incoming chunk server every time would be tedious, we can have the chunk server be aware of its assigned id number and provide that in the message. Then just use that id number as the index in this vector.

If we can't answer that now I'd suggest we leave this as it is until we can. Looping through a vector will not slow our tests down during development anyway since we'll have at most 2 or 3 chunk servers

src/sadmd/sadmd.cpp
damanmkumar: `touch` actually updates the last accessed time of a file. You can see it with `$ ls -l file`. So, the second `touch` technically does modify the file. What happens when a client requests a create operation on a file that already exists should ideally be a higher level decision. By having `create_file` take the decision to silently let this happen, we're giving it more responsibility than it should have.

`create_file` now has two responsibilities. Creating a file, and deciding what happens when a file exists already.

include/sadfs/sadmd/sadmd.hpp
damanmkumar: Yes! I'm talking about `sqlite3* const files_db_` and not `sqlite3 const* files_db_`

include/sadfs/sadmd/sadmd.hpp
damanmkumar: Yeah, the chunk server will need to identify itself when it sends a heartbeat message. I can think of two ways we can implement this.

The simplest approach is to have the chunk server send its ID, which we can use as a key. Using its IP address to identify it will not be a good idea from a correctness perspective. A crashed chunk server might get a new IP address on recovery. Hashing a `std::string` will introduce unnecessary overhead at runtime. So, a unique ID is the better approach here.

Another way we could do it is assign IDs to chunk servers starting from 0, in the order they contact the master server. This will let us use the ID as an index, but it is extra work that the master server has to do.

I suggest we design these things keeping our use cases in mind. At least as much as possible. Otherwise, we'll end up refactoring a lot of things once we start implementing higher level functionality.

abmono93: > it is extra work that the master server has to do

It doesn't seem like a lot of extra work to do something like
`id=active_chunk_servers.size()` and then `active_chunk_servers.emplace_back()`
in terms of actual computation I would guess it is less than the extra overhead of a hash lookup for each heartbeat

The only alternative I see is `std::unorderd_map<unsigned int, chunk_server_info>`  . Do you think that'd be better? Or do you have something else in mind?

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,60 @@
...
 
+#include <chrono>
 #include <string>
+#include <sqlite3.h>
+#include <unordered_map>
+#include <vector>
 
 namespace sadfs {
 
+using chunkid = uint64_t;
```
damanmkumar, line 14: Let's call this uuid, since we'll need to use it for identifying chunks, chunk servers, and any other things we might need to uniquely identify.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,60 @@
...
+
+// all the information needed about a chunk server
+struct chunk_server_info{
+	inet::service service;
+	uint64_t total_space;
```
damanmkumar: `{total|used}_space` will need to be renamed since we're encoding the number of chunks stored. Maybe `max_chunks` and `stored_chunks` or anything else you might deem fit.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,60 @@
...
+	uint64_t used_space;
+	int ttl;
```
damanmkumar, line 22: `ttl` should not be a signed integer, right?

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,60 @@
...
+};
+
+// all the information needed about a file
+struct file_info{
+	int ttl;
+	std::vector<chunkid> chunkids;
+	// decides whether file is currently locked for writing
+	bool locked();
```
damanmkumar: Since we aren't implementing `locked()` yet, let's remove it now and introduce it when we implement it.

include/sadfs/sadmd/sadmd.hpp
damanmkumar: Actually, I'm not opposed to `chunkid` since it is more readable.

abmono93: The motive behind this was to differentiate when I'm talking about chunkids vs. some other unsigned integer. I just feel like chunkids show up all over the place  - they're used in file_info, as keys in the chunkids_ map, stored in the database, and will be passed around in almost every read/write control message - so it makes sense to give them an easily identifiable name. Chunk server ids are used in a more specific context (they are basically there for the master to keep track of them in the context of a single data structure) so I didn't think the same thing applied

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,60 @@
...
+	void create_file(std::string const&);
 private:
 	// reads the message from a socket that just received some data
 	std::string process_message(sadfs::socket const&);
+	// loads file metadata from disk
+	void load_files();
+	// runs an sql statement on system_files_
+	void db_command(std::string const&) const noexcept;
+
+	const inet::service service_;
```
damanmkumar: Let's stick to the "east const" notation, which will make this `inet::service const service_;` to be consistent across our code base.

include/sadfs/sadmd/sadmd.hpp
abmono93: Yes `*_chunks` is much clearer

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -3,22 +3,60 @@
...
+	// in memory representation of each file
+	std::unordered_map<std::string, file_info> files_;
+	// list of active chunk servers
+	std::unordered_map<uint64_t, chunk_server_info> chunk_servers_;
+	// map from chunkid to list of chunk servers
+	std::unordered_map<chunkid, std::vector<chunk_server_info*> > chunkids_;
```
damanmkumar: `chunkids_` sounds like a collection of `chunkid`s, right? And this is more of a collection of chunk servers.

include/sadfs/sadmd/sadmd.hpp
abmono93: I switched from `time_t` to  `int` because you said above it can be represented in a regular integer. I've never used chrono so I was just taking your word for it haha. 

You think it should be `unsigned int`?

include/sadfs/sadmd/sadmd.hpp
damanmkumar: Maybe we can rename the current `chunk_servers_` data structure to `chunk_server_metadata`, and `chunkids_` to `chunk_servers`. That way, this is what our code will look like:
```
chunk_server_metadata[server_id] // refers to metadata belonging to chunk server specified by id
chunk_servers[chunk_id]          // refers to a list of servers that host chunk_id
```

abmono93: I guess it is kind of like a dictionary with one entry for each `chunkid` that tells you where you can find that chunk. And the most straightforward way to get a list of all chunks in the system would be to iterate through this map. Whereas `chunk_servers_` is the authoritative list of chunk servers. But if there is a more obvious naming convention I am open to it. maybe `chunk_locations_` ?

(This was in response to your first message before I saw the second one)

include/sadfs/sadmd/sadmd.hpp
damanmkumar: Ah, sorry! It should definitely be `std::chrono::steady_clock::duration::rep` but aliased as something more readable. Let's leave it as an `int` for now. We can use a more suitable type once we implement TTL. I'll try to quickly put something together tonight.

src/sadmd/sadmd.cpp
```diff
@@ -63,4 +83,51 @@ process_message(sadfs::socket const& sock)
...
+
+void sadmd::
+load_files()
+{
+	std::cout << "Files loaded:\n";
+	auto res = 0;
+	sqlite3_stmt* stmt;
+	if (sqlite3_prepare_v2(files_db_, "SELECT * FROM files;", -1, &stmt, NULL) != 0)
+	{
+			std::cerr << sqlite3_errmsg(files_db_) << '\n';
```
damanmkumar: This needs to be formatted. Looks like there's an extra tab.

src/sadmd/sadmd.cpp
```diff
@@ -63,4 +83,51 @@ process_message(sadfs::socket const& sock)
...
+	}
+
+	while ((res = sqlite3_step(stmt)) == SQLITE_ROW)
+	{
+		auto filename = std::string{reinterpret_cast<const char*>(sqlite3_column_text(stmt, 0))};
```
damanmkumar: Is this line longer than 79 characters? Lets stick to that limit. It's a standard of sorts. Super long lines get wrapped to another line in my editor.

src/sadmd/sadmd.cpp
```diff
@@ -63,4 +83,51 @@ process_message(sadfs::socket const& sock)
...
+		const unsigned char* cids = sqlite3_column_text(stmt, 1);
+		std::cout << filename << ": " << cids << '\n';
+		create_file(filename);
+		//TODO: convert cids into a vector of ints and add to file_info object
+	}
+	sqlite3_finalize(stmt);
+}
+
+sqlite3* sadmd::
+open_db()
```
damanmkumar: Lets also move all functions that need not be exposed to code in other `.cpp` files into `unnamed namespace`s. The compiler can perform some really cool optimizations that way.

abmono93: That would apply to most of the functions in this file, right? Should I just wrap the whole class in an unnamed namespace?

damanmkumar: If you don't see the need in a function being linked to from another `.cpp` file, throw it into an unnamed namespace. In fact, `open_db` is a utility function, right? There's no reason it should be a member function of `sadmd`. So, you can remove it from the private section of the class in `sadmd.hpp` as well. If a function isn't using a class's member variables/functions, let's just make it a utility function and keep the interface clean.

abmono93: Wait sorry.. that would apply to all private functions though right? i.e. `process_message`

src/sadmd/sadmd.cpp
```diff
@@ -63,4 +83,51 @@ process_message(sadfs::socket const& sock)
 	return result;
 }
 
+void sadmd::
+db_command(std::string const& stmt) const noexcept
+{
+	if (sqlite3_exec(files_db_, stmt.c_str(), NULL, NULL, NULL) != 0)
+	{
+		std::cerr << "Error running {" << stmt << "}: ";
```
damanmkumar, line 109: We'll need to handle this error as well.

damanmkumar (approved):

PR #30, opened 2020-11-13
Created sadfuse class and sadfsd-bootstrap

### LINKED ISSUE #5: Start-up configuration - File System daemon
As an admin, I should be able to configure default options for the File System daemon in a configuration file, and override them over the command line

### LINKED ISSUE #35: Add libfuse support
User should be able to mount file system.

shu-g:
* Implemented a bare bones fuse wrapper
* Modified CMakeFile to support libfuse
* Added sadfsd-bootstrap file


doc/ideas.md
```diff
@@ -98,6 +108,41 @@ file descriptors.
...
+			}
+			
+		Inode : {
+				char[64] Filename, ino_t Inode number;
+				char[64] Filename, ino_t Inode number;
+				...
+			}
+		...
+	}
+	Metadata Content Table
```
abmono93, line 128: Are you thinking we will have a table like this in addition to the ones described below?

abmono93 (approved):

doc/ideas.md
shu-g: Yeah, I think we can have Directory Content Table, Metadata Content Table, Chunks Table and Chunk Servers Table.
From an inode number in query, we can lookup files in that directory in the DC table.
For more info about file, we can lookup inode in MC table.
If we get read request, from the MC table, we'll get mapping to chunks in the C table and CS table.

src/CMakeLists.txt
```diff
@@ -63,6 +63,26 @@ foreach (TARGET sadmd sadcd) # sadfsd)
 		PRIVATE ${Boost_LIBRARIES})
 endforeach()
 
+# find fuse library. this sets the FUSE3_INCLUDE_DIRS and FUSE3_LIBRARIES
+# variables
+find_package(PkgConfig REQUIRED)
+pkg_check_modules(FUSE REQUIRED fuse)
+
+add_executable (sadfsd-bootstrap
```
damanmkumar: Adding the executable and letting CMake know that Boost's program_options should be linked should be done in the foreach loop like  we do for `sadmd` and `sadcd`. Any sadfsd-specific configuration can be done separately.

src/CMakeLists.txt
```diff
@@ -63,6 +63,26 @@ foreach (TARGET sadmd sadcd) # sadfsd)
...
+    ${COMMON_SOURCE_FILES})
+target_include_directories (sadfsd-bootstrap
+    PRIVATE ${Boost_INCLUDE_DIRS}
+    SYSTEM PRIVATE ${FUSE_INCLUDE_DIRS})
+target_link_libraries (sadfsd-bootstrap
+    PRIVATE sadfs_protobuf
+    PRIVATE ${Boost_LIBRARIES}
+    PRIVATE ${FUSE_LINK_LIBRARIES})
+target_compile_options(sadfsd-bootstrap
+    PRIVATE ${FUSE_CFLAGS})
```
damanmkumar, line 90: Not a review comment. What is the significance of `FUSE_CFLAGS` here?

include/sadfs/sadfsd/sadfuse.hpp
```diff
@@ -0,0 +1,65 @@
...
+
+class sadfuse
+{
+public:
+    // delete copy constructor and copy assignment operator
+    sadfuse(sadfuse const&) = delete;
+    sadfuse &operator=(sadfuse const&) = delete;
+
+    // returns singleton instance of the class
+    static sadfuse& instance();
```
damanmkumar: Why does `sadfuse` need to be a singleton class? It's purpose is to "set up" fuse, right?

include/sadfs/sadfsd/sadfuse.hpp
```diff
@@ -0,0 +1,65 @@
...
+    
+    // creates connection to master at ip and port and stores it in
+    // master_socket_
+    void set_master_connection(char const* ip, int port);
+
+    // starts fuse_main()
+    int run(int argc, char** argv);
+
+    // returns operations_
+    fuse_operations& operations();
```
damanmkumar: If `load_operations()` sets up `operations_`, why does `sadfuse` need to publicly expose it?

include/sadfs/sadfsd/sadfuse.hpp
```diff
@@ -0,0 +1,65 @@
...
+    // open a file
+    static int open(const char* path, fuse_file_info* fi);
+
+    // read data from an open file
+    static int read(const char* path, char* buf, size_t size, off_t offset,
+             fuse_file_info* fi);
+
+    fuse_operations operations_;
+    
+    std::shared_ptr<sadfs::socket> master_socket_;
```
damanmkumar: Our master server isn't designed to handle long standing connections with clients. Every request must be a separate connection.

What is the point of a `shared_pointer` here?

src/sadfsd/sadfsd-bootstrap.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+	// read options not specified via the CLI
+	sadfs::bootstrap::parse_config_file(variables, options);
+
+	// verify that mandatory options have been configured
+	// po::notify throws if they are not
+	sadfs::bootstrap::verify(variables);
+
+    // store reference to the singleton sadfuse object
+    sadfs::sadfuse& sadfuse = sadfs::sadfuse::instance();
```
damanmkumar: The file needs to be formatted correctly.

src/sadfsd/sadfsd-bootstrap.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+    // establish connection with master
+    sadfuse.set_master_connection(
+            variables["ipaddress"].as<std::string>().data(),
+            variables["port"].as<std::uint16_t>());
+    
+    // since the first argument is skipped by fuse_main, it is initialized as
+    // empty string
+    std::array<std::string, 2> fuse_args = std::array
```
damanmkumar: We're using the `auto <name> = <type>{...};` syntax throughout the code base. So, this needs to change, along with `sadfuse`.

src/sadfsd/sadfsd-bootstrap.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+	// allocate space for pointers, including terminating nullptr
+	auto fuse_argv = std::array<char*, fuse_args.size() + 1>{nullptr};
+
+	// construct fuse_argv from fuse_args
+	auto get_ptr = [](auto& str) { return str.data(); };
+	std::transform(fuse_args.begin(), fuse_args.end(), fuse_argv.begin(),
+                   get_ptr);
+
+    int status = sadfuse.run(fuse_args.size(), fuse_argv.data());
```
damanmkumar: This should be `auto status...` as well

include/sadfs/sadfsd/sadfuse.hpp
```diff
@@ -0,0 +1,65 @@
...
+private:
+    // private constructor to enforce singleton property
+    sadfuse();
+
+    // assign file system operation function pointers to the corresponding
+    // file system operation definitions
+    void load_operations();
+    
+    // get file attributes
+    static int getattr(const char* path, struct stat* stbuf);
```
damanmkumar: We're using the "east const" notation in our code base. So, this would be `char const*`.

src/sadfsd/sadfuse.cpp
```diff
@@ -0,0 +1,120 @@
...
+sadfuse()
+{
+    memset(&operations_, 0, sizeof operations_);
+    load_operations();
+}
+
+void sadfuse::
+load_operations()
+{
+    operations_.getattr = &getattr;
```
damanmkumar: This is more of a nit. Formatting this like
```
    operations_.getattr  = &getattr;
    operations_.readlink = nullptr;
    operations_.mknod    = nullptr;
    operations_.mkdir    = nullptr;
```
will be easier to read and parse.

src/sadfsd/sadfuse.cpp
```diff
@@ -0,0 +1,120 @@
...
+fuse_operations& sadfuse::
+operations()
+{
+    return operations_;
+}
+
+sadfuse::
+sadfuse()
+{
+    memset(&operations_, 0, sizeof operations_);
```
damanmkumar: There's no reason to call `memset` here. The struct is zero-initialized if `sadfuse`'s constructor is called.

src/sadfsd/sadfuse.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+// standard includes
+#include <cstring> // memset()
+
+namespace sadfs {
+
+sadfuse& sadfuse::
+instance()
+{
+    static sadfuse instance;
```
damanmkumar: This syntax will not initialize the `operations_` member variable of `sadfuse`. Use `auto instance = <type>{};` to avoid calling `memset`.

src/sadfsd/sadfuse.cpp
```diff
@@ -0,0 +1,120 @@
...
+set_master_connection(char const* ip, int port)
+{
+    master_socket_ = std::make_shared<sadfs::socket>(
+                        inet::service(ip, port).connect());
+}
+
+int sadfuse::
+run(int argc, char** argv)
+{
+    return fuse_main(argc, argv, &operations(), this);
```
damanmkumar: Why not `&operations_`? That is what the compiler will do anyway.

damanmkumar (changes requested):
made a few comments

src/CMakeLists.txt
shu-g: The compile flags for fuse contain path to include directories.

include/sadfs/sadfsd/sadfuse.hpp
shu-g: Initial thought was that I should use global functions for the set up or create class with single instance. But now, I've modified the program so that this is normal class and made operations_ a static member.

include/sadfs/sadfsd/sadfuse.hpp
shu-g: You're right, changed it.

include/sadfs/sadfsd/sadfuse.hpp
shu-g: Since there is no empty constructor for socket and wasn't initializing in sadfuse constructor, had to use shared_pointer.
Modified it to be an object of inet::service.

src/sadfsd/sadfuse.cpp
shu-g: You're right.

src/CMakeLists.txt
shu-g: Had to move add_dependency out of the loop, let me know if more changes are required.

src/CMakeLists.txt
damanmkumar: Isn't that handled by `FUSE_INCLUDE_DIRS`?

include/sadfs/sadfsd/sadfuse.hpp
```diff
@@ -19,45 +19,35 @@ class sadfuse
...
     // read data from an open file
-    static int read(const char* path, char* buf, size_t size, off_t offset,
+    static int read(char const* path, char* buf, size_t size, off_t offset,
              fuse_file_info* fi);
 
-    fuse_operations operations_;
+    static fuse_operations operations_;
     
-    std::shared_ptr<sadfs::socket> master_socket_;
+    inet::service master_service_;
```
damanmkumar: If our `read`, `open`, and other `static` functions need to access `master_service_`, it needs to be `static` as well.

include/sadfs/sadfsd/sadfuse.hpp
```diff
@@ -0,0 +1,55 @@
+#ifndef SADFS_SADFSD_SADFUSE_HPP
+#define SADFS_SADFSD_SADFUSE_HPP
+#include <sadfs/comm/inet.hpp>
+
+#include <memory> // std::shared_ptr
```
damanmkumar: This include will no longer be required since we do not need a `std::shared_ptr`

include/sadfs/sadfsd/sadfuse.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#ifndef FUSE_USE_VERSION
+#define FUSE_USE_VERSION 26
+#endif
+#define _FILE_OFFSET_BITS 64
+#include <fuse.h>
+
+namespace sadfs {
+
+class sadfuse
```
damanmkumar: The purpose of this class is to set up FUSE, but what it really is doing is setting up the file system on the client machine, right? The name `sadfuse` is kinda exposing an implementation detail (that we're using FUSE). ~I feel `sadfs-bootstrap` is a better-suited name.~ We need a different name. Thoughts? @abmono93 , what do you think?

src/sadfsd/sadfsd-bootstrap.cpp
```diff
@@ -0,0 +1,119 @@
...
+	// read options not specified via the CLI
+	sadfs::bootstrap::parse_config_file(variables, options);
+
+	// verify that mandatory options have been configured
+	// po::notify throws if they are not
+	sadfs::bootstrap::verify(variables);
+
+    // create object of sadfuse class
+    auto sadfuse = sadfs::sadfuse{
+            variables["ipaddress"].as<std::string>().data(),
```
damanmkumar: `variables["ipaddress"].as<std::string>().c_str()` might express the intention better.

src/sadfsd/sadfsd-bootstrap.cpp
```diff
@@ -0,0 +1,119 @@
...
+	{
+        std::string(),
+		variables["mountpoint"].as<std::string>()
+	};
+
+	// allocate space for pointers, including terminating nullptr
+	auto fuse_argv = std::array<char*, fuse_args.size() + 1>{nullptr};
+
+	// construct fuse_argv from fuse_args
+	auto get_ptr = [](auto& str) { return str.data(); };
```
damanmkumar, line 114: `str.c_str()` expresses the idea better. We should correct this in `sadmd` and `sadcd` as well.

include/sadfs/sadfsd/sadfuse.hpp
abmono93: This is basically the client process's analogue of of `sadmd` and `sadcd` right? So how about `sadfsd`?

src/sadfsd/sadfsd-bootstrap.cpp
```diff
@@ -0,0 +1,119 @@
...
+	std::transform(fuse_args.begin(), fuse_args.end(), fuse_argv.begin(),
+                   get_ptr);
+
+    auto status = sadfuse.run(fuse_args.size(), fuse_argv.data());
```
abmono93: Shouldn't run() be called from a process that is spawned by the bootstrap process? The way this is setup there will be an executable called sadfsd-bootstrap running as long as the user is running our filesystem, right?

include/sadfs/sadfsd/sadfuse.hpp
damanmkumar: The problem here is that the `d` stands for daemon, and this class isn't a daemon.

src/sadfsd/sadfsd-bootstrap.cpp
damanmkumar: No, fuse runs a different way. The `run()` function will basically tell the kernel/fuse which functions it/they should call for each operation (read, write, ...), and return immediately. At this point the remote file system is "mounted."

Any subsequent calls to these operations within that mounted file system will be redirected to our functions, which are the static functions in this class.

src/sadfsd/sadfuse.cpp
```diff
@@ -0,0 +1,102 @@
...
+sadfuse::
+sadfuse(char const* ip, int port) : master_service_(ip, port)
+{
+    load_operations();
+}
+
+int sadfuse::
+run(int argc, char** argv)
+{
+    return fuse_main(argc, argv, &operations_, this);
```
abmono93: Where is fuse_main defined?

include/sadfs/sadfsd/sadfuse.hpp
abmono93: Right. How about something like `filesys` or `sadfilesys` ?

src/sadfsd/sadfuse.cpp
damanmkumar: It's defined in `fuse.h`. This brings up a good thing. We should definitely be calling this function (and all functions we haven't written) as `::fuse_main(...)`

damanmkumar: Question for @shu-g : Why are we passing in the `this` pointer here?

include/sadfs/sadfsd/sadfuse.hpp
damanmkumar: Either of those should work for now. Also, I just realised a problem here, and raised issue #42

src/CMakeLists.txt
shu-g: I thought so too, but was getting linking errors on macOS without them.

include/sadfs/sadfsd/sadfuse.hpp
shu-g: Using static shared pointer to master service. Assignment of inet::service was failing because the member ip_ and port_ are declared const.

src/sadfsd/sadfsd-bootstrap.cpp
shu-g: Done

src/sadfsd/sadfsd-bootstrap.cpp
shu-g: Done

src/sadfsd/sadfuse.cpp
shu-g: 4th parameter is used to pass private data that gets stored in fuse context, in our case this is used to pass master_service_ and operations_

include/sadfs/sadfsd/sadfuse.hpp
shu-g: Lets do sadfilesys then

src/sadfsd/sadfsd-bootstrap.cpp
shu-g: C++17 onwards data() returns char*, c_str() returns char const*.
Since fuse_main expects char*, we need to use data()

include/sadfs/sadfsd/sadfilesys.hpp
```diff
@@ -0,0 +1,58 @@
...
+public:
+    // delete copy constructor and copy assignment operator
+    sadfilesys(sadfilesys const&) = delete;
+    sadfilesys &operator=(sadfilesys const&) = delete;
+    
+    // contructor
+    sadfilesys(char const* ip, int port);
+
+    // starts fuse_main()
+    int run(int argc, char** argv);
```
damanmkumar: `run` isn't an appropriate name in this context. `fs.run()` does not read right. How about `bootstrap`?

include/sadfs/sadfsd/sadfilesys.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+private:
+    // assign file system operation function pointers to the corresponding
+    // file system operation definitions
+    void load_operations();
+
+    // returns static pointer to object of this class
```
damanmkumar: This comment doesn't explain what this method really does. We should mention that it gets the pointer from `fuse_context()`

include/sadfs/sadfsd/sadfilesys.hpp
```diff
@@ -0,0 +1,58 @@
...
+                fuse_file_info* fi);
+
+    // open a file
+    int open(char const* path, fuse_file_info* fi);
+
+    // read data from an open file
+    int read(char const* path, char* buf, size_t size, off_t offset,
+             fuse_file_info* fi);
+
+    static fuse_operations operations_;
```
damanmkumar: This no longer needs to be `static` right?

include/sadfs/sadfsd/sadfilesys.hpp
```diff
@@ -0,0 +1,58 @@
+#ifndef SADFS_SADFSD_SADFILESYS_HPP
+#define SADFS_SADFSD_SADFILESYS_HPP
+#include <sadfs/comm/inet.hpp>
+
+#include <memory> // std::shared_ptr
```
damanmkumar: This header is not required anymore

include/sadfs/sadfsd/sadfuse.hpp
```diff
@@ -0,0 +1,55 @@
...
+             fuse_file_info* fi);
+
+    static fuse_operations operations_;
+    
+    inet::service master_service_;
+};
+
+} // sadfs namespace;
+
+#endif // SADFS_SADFSD_SADFUSE_HPP
```
damanmkumar: This file should be deleted

src/sadfsd/sadfilesys.cpp
```diff
@@ -0,0 +1,126 @@
+/* Definitions for the sadfilesys class */
+
+// sadfs-specific includes
+#include <sadfs/comm/inet.hpp>
+#include <sadfs/sadfsd/sadfilesys.hpp>
+
+// standard includes
+#include <cstring> // memset()
```
damanmkumar: No longer required, correct?

src/sadfsd/sadfilesys.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // TODO
+    return -ENOENT;
+}
+
+int sadfilesys::
+readdir(char const* path, void* buf, fuse_fill_dir_t filler, off_t off,
+        fuse_file_info* fi)
+{
+    // TODO
+    master_service_.connect();
```
damanmkumar: If this statement is here for testing, we can add this line while testing on dev machines. Let's not push it to github.

src/sadfsd/sadfsd-bootstrap.cpp
damanmkumar: I wonder why a lot of these interfaces expect non-const pointers when they only read from them

src/sadfsd/sadfsd-bootstrap.cpp
```diff
@@ -0,0 +1,121 @@
...
+            variables["ipaddress"].as<std::string>().c_str(),
+            variables["port"].as<std::uint16_t>()};
+
+    // since the first argument is skipped by fuse_main, it is initialized as
+    // empty string
+    auto filesys_args = std::array
+	{
+        std::string(),
+		variables["mountpoint"].as<std::string>()
+	};
```
damanmkumar: The formatting is off, here

src/sadfsd/sadfuse.cpp
```diff
@@ -0,0 +1,102 @@
+/* Definitions for the sadfuse class */
```
damanmkumar: This file needs to go as well!

damanmkumar (changes requested):
A few minor comments

src/sadfsd/sadfuse.cpp
shu-g: Yup

src/sadfsd/sadfsd-bootstrap.cpp
shu-g: Oh, looks fine on my editor.

src/sadfsd/sadfilesys.cpp
shu-g: Ok.

src/sadfsd/sadfilesys.cpp
shu-g: Oh right.

include/sadfs/sadfsd/sadfilesys.hpp
shu-g: Oh right.

include/sadfs/sadfsd/sadfilesys.hpp
shu-g: Oh right.

include/sadfs/sadfsd/sadfilesys.hpp
shu-g: modified it to:
retrieves instance of this class with the help of fuse_get_context() and returns a static pointer to it

include/sadfs/sadfsd/sadfilesys.hpp
shu-g: Yeah works

include/sadfs/sadfsd/sadfuse.hpp
shu-g: Yup

src/CMakeLists.txt
```diff
@@ -53,21 +50,45 @@ find_package (Boost
...
 		PRIVATE ${Boost_LIBRARIES})
 endforeach()
 
+# for sadmd and sadcd, do the following
+# 1. mark their bootstrap targets as dependencies
+# 2. mark sadfs_protobuf as a dependency
+foreach(TARGET sadmd sadcd)
+	add_dependencies (${TARGET} ${TARGET}-bootstrap)
+	target_link_libraries (${TARGET} PRIVATE sadfs_protobuf)
+endforeach()
```
shu-g, line 73: @damanmkumar , sorry, I should've asked this in the patch-2 review but anyway:
Why should TARGET-bootstrap be built before TARGET?
In line #59, why does BOOTSTRAP_TARGET need to be linked against sadfs_protobuf?

damanmkumar: We add it as a dependency only for the sake of convenience. Since we need the bootstrap targets to start our main server processes, it makes sense to also build `sadmd-bootstrap` when you run `make sadmd`. There isn't any dependency code-wise!

Linking the bootstrap targets against `sadfs_protobuf` is just something that used to be relevant, but no longer is. The original idea was to get the bootstrap process to pass some information to the main process using protobufs. It can be safely removed now. Feel free to either do it now, or raise an issue -- we can take care of it later.

damanmkumar (approved):
LGTM

PR #34, opened 2020-11-18
Maintain db

### LINKED ISSUE #33: Maintain database
The master server should update the database once all time-critical functions (i.e. responding to requests) are completed. This includes:

- Creating/removing a new file

- Adding/removing a chunkid to a file

Alternatively, the master server could write all the contents of the databse on shutdown.

abmono93:
Created a function for the master server to save all files to the database. This is intended to be called on shutdown.

As part of this I also changed create_file() so that it can load the saved list of chunkids into a vector.

Closes #33

### DIFF
--- a/src/sadmd/sadmd.cpp
+++ b/src/sadmd/sadmd.cpp
@@ -14,10 +14,15 @@
 #include <sstream>  // std::osstream
 #include <unistd.h> // read/write
 
-#define FILENAME_COL 0
-#define CHUNKID_STR_COL 1
-
 namespace {
+
+namespace constants {
+
+// columns in our SQLite database
+constexpr auto filename_col = 0;
+constexpr auto chunkid_str_col = 1;
+
+} // (local) constants namespace
 
 sqlite3*
 open_db()
@@ -196,9 +201,9 @@ load_files()
 	while (sqlite3_step(stmt) == SQLITE_ROW)
 	{
 		auto filename = std::string{reinterpret_cast<const char*>(
-			sqlite3_column_text(stmt, FILENAME_COL))};
+			sqlite3_column_text(stmt, constants::filename_col))};
 		auto cids = std::string{reinterpret_cast<const char*>(
-			sqlite3_column_text(stmt, CHUNKID_STR_COL))};
+			sqlite3_column_text(stmt, constants::chunkid_str_col))};
 		create_file(filename, cids);
 	}
 	sqlite3_finalize(stmt);
--- a/src/sadmd/sadmd.cpp
+++ b/src/sadmd/sadmd.cpp
@@ -8,6 +8,7 @@
 #include <array>
 #include <cstring>  // std::strerror
 #include <iostream>
+#include <iterator>
 #include <string>
 #include <sqlite3.h>
 #include <sstream>  // std::osstream
--- a/src/sadmd/sadmd.cpp
+++ b/src/sadmd/sadmd.cpp
@@ -145,7 +145,7 @@ create_file(std::string const& filename, std::string const& existing_chunks)
 {
 	if (!files_.count(filename))
 	{
-		auto info = file_info{0, };
+		auto info = file_info{};
 		parse_chunkid_str(info.chunkids, existing_chunks);
 		files_.emplace(filename, info);
 	}
--- a/src/sadmd/sadmd.cpp
+++ b/src/sadmd/sadmd.cpp
@@ -10,6 +10,7 @@
 #include <iostream>
 #include <string>
 #include <sqlite3.h>
+#include <sstream>  // std::osstream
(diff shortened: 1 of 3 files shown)


src/sadmd/sadmd.cpp
```diff
@@ -127,12 +194,34 @@ load_files()
 	{
 		auto filename = std::string{reinterpret_cast<const char*>(
 			sqlite3_column_text(stmt, 0))};
-		const unsigned char* cids = sqlite3_column_text(stmt, 1);
-		std::cout << filename << ": " << cids << '\n';
-		create_file(filename);
-		//TODO: convert cids into a vector of ints and add to file_info object
+		auto cids = std::string{reinterpret_cast<const char*>(
+			sqlite3_column_text(stmt, 1))};
```
damanmkumar: The magic number `1` here should be replaced by something more readable. Same with the `0` on line 196.

src/sadmd/sadmd.cpp
```diff
@@ -127,12 +194,34 @@ load_files()
...
+void sadmd::
+save_files() const noexcept
+{
+	for (auto file : files_)
+	{
+		auto sql_command = std::string{};
+		auto filename = std::string{file.first};
+		auto chunkids = chunkid_str(file.second.chunkids);
+
+		if(db_contains(file.first))
```
damanmkumar, line 221: Not really a review comment for the time being: If possible, we should come up with a strategy that avoids checking if every file exists. This will not scale as our file count increases. Maybe we should add an issue, and look at it if we're interested in polishing our code after the quarter ends.

src/sadmd/sadmd.cpp
```diff
@@ -86,11 +133,13 @@ start()
 }
 
 void sadmd::
-create_file(std::string const& filename)
+create_file(std::string const& filename, std::string const& existing_chunks)
 {
 	if (!files_.count(filename))
 	{
-		files_.emplace(filename, file_info{0, });
+		auto info = file_info{0, };
```
damanmkumar: `file_info{}` works as well. There's no reason to specify the `0` here, is there?

src/sadmd/sadmd.cpp
```diff
@@ -112,10 +161,28 @@ db_command(std::string const& stmt) const noexcept
...
 
+bool sadmd::
+db_contains(std::string const& filename) const
+{
+	auto contains = false;
+	sqlite3_stmt* stmt;
+	auto cmd = std::string{"SELECT * FROM files WHERE filename='" + filename +
+		"';"};
+
+	if (sqlite3_prepare_v2(files_db_, cmd.c_str(), -1, &stmt, NULL) != 0)
```
damanmkumar: Can we replace `NULL`s with `nullptr`?

damanmkumar (changes requested):
a few minor comments

src/sadmd/sadmd.cpp
abmono93: Agreed. The better strategy here will be to have two lists, one of new files and one of updated files. Then the master server would periodically do some garbage collection and only update the files on these two lists. But then we'd run the risk of creating bugs every time we change virtually anything about how the master server works. So it makes more sense to switch to this strategy after that's mostly implemented anyway.

src/sadmd/sadmd.cpp
abmono93: The ttl should be set to zero for a new file. Will it default to zero? I'm still relatively new to the `auto = ..{}` syntax

src/sadmd/sadmd.cpp
damanmkumar: That's an interesting approach. We would initialize the existing files list on start up, and all new files go into the new files list. While serializing, we know which files needs to be "created," and which need to be updated.

I think a better way to implement this would be to write to the database every time a client wants to create a new file. Currently, if a client wants to create a file, we will happily add it to our in-memory data structure, and tell the client everything went ok. However, if the master crashes, this data is lost. So, ideally, file info should go to disk before we inform the client that the create request succeeded.

damanmkumar: We can actually club the two approaches. Every newly created file goes to a "temporary" database on disk. On shutdown (and start up) we simply move all data from the temporary database into the actual one. That way, file create requests can be served quickly (inserting into a smaller database should be faster?), and we achieve correctness. In case of crashes, the server will look for this temporary database and move it into the main database upon start up.

src/sadmd/sadmd.cpp
damanmkumar: Yup, `{}` will default initialize things. if we were to use `file_info info;`, primitive data types like `int`, `char`, and `float` will not be initialized.

PR #41, opened 2020-11-23
Implemented internal protobuf for db serialization

### LINKED ISSUE #40: Replace serialization with internal protobufs

abmono93:
Closes #40

### DIFF
--- a/src/comm/internal.proto
+++ b/src/comm/internal.proto
@@ -1,0 +1,6 @@
+syntax = "proto3";
+package sadfs.proto.internal;
+
+message chunkids {
+    repeated uint64 chunkid = 1;
+}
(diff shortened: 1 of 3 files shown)


src/comm/internal.proto
```diff
@@ -0,0 +1,6 @@
+syntax = "proto3";
+package sadfs.proto.internal;
+
+message chunkids {
+    repeated uint64 chunkid = 1;
```
damanmkumar, line 5: Since this is a repeated field, its name should be a plural noun. `chunkids` is a good option, in which case we should change the name of the message.

abmono93: How about `chunkids_msg`?

damanmkumar (approved):
Looks good

PR #43, opened 2020-11-25
Created functions to maintain list of servers

### LINKED ISSUE #31: Master functions to maintain list of chunk servers
The master server should be able to add or remove chunk servers to the network and update the ttl of one in response to a heartbeat

abmono93:
Closes #31

### DIFF
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -59,7 +59,7 @@ private:
 	// functions for maintaining chunk servers 
 
 	// returns true on success
-	bool add_server_to_network(serverid, char const*, int, uint64_t);
+	bool add_server_to_network(serverid, char const*, int, uint64_t, uint64_t);
 	void remove_server_from_network(serverid) noexcept;
 	void register_server_heartbeat(serverid) noexcept;
 	bool is_active(serverid) const noexcept;
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -21,7 +21,7 @@ struct chunk_server_info
 	inet::service service;
 	uint64_t max_chunks;
 	uint64_t chunk_count;
-	time_point ttl;
+	time_point expiration_point;
 };
 
 // all the information needed about a file
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -13,7 +13,7 @@ namespace sadfs {
 
 using chunkid = uint64_t;
 using serverid = uint64_t;
-using tick = std::chrono::steady_clock::rep;
+using time_point = std::chrono::steady_clock::rep;
 
 // all the information needed about a chunk server
 struct chunk_server_info
@@ -21,7 +21,7 @@ struct chunk_server_info
 	inet::service service;
 	uint64_t max_chunks;
 	uint64_t chunk_count;
-	tick ttl;
+	time_point ttl;
 };
 
 // all the information needed about a file
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -13,7 +13,7 @@ namespace sadfs {
 
 using chunkid = uint64_t;
 using serverid = uint64_t;
-using nanosec = std::chrono::steady_clock::duration::rep;
+using ticks = std::chrono::steady_clock::rep;
 
 // all the information needed about a chunk server
 struct chunk_server_info
@@ -21,7 +21,7 @@ struct chunk_server_info
 	inet::service service;
 	uint64_t max_chunks;
 	uint64_t chunk_count;
-	nanosec ttl;
+	ticks ttl;
 };
 
 // all the information needed about a file
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -60,9 +60,9 @@ private:
 
 	// returns true on success
 	bool add_server_to_network(serverid, char const*, int, uint64_t);
-	void remove_server_from_network(serverid);
-	void register_server_heartbeat(serverid);
-	bool is_active(serverid);
+	void remove_server_from_network(serverid) noexcept;
+	void register_server_heartbeat(serverid) noexcept;
+	bool is_active(serverid) const noexcept;
 
 	inet::service const service_;
 	// in memory representation of each file
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -12,6 +12,8 @@
 namespace sadfs {
 
 using chunkid = uint64_t;
+using serverid = uint64_t;
+using nanosec = std::chrono::steady_clock::duration::rep;
 
 // all the information needed about a chunk server
 struct chunk_server_info
@@ -19,7 +21,7 @@ struct chunk_server_info
 	inet::service service;
 	uint64_t max_chunks;
 	uint64_t chunk_count;
-	int ttl;
+	nanosec ttl;
 };
 
 // all the information needed about a file
@@ -54,11 +56,19 @@ private:
 	// returns true if the database contains a file with the given name
 	bool db_contains(std::string const&) const;
 
+	// functions for maintaining chunk servers 
+
+	// returns true on success
+	bool add_server_to_network(serverid, char const*, int, uint64_t);
+	void remove_server_from_network(serverid);
+	void register_server_heartbeat(serverid);
+	bool is_active(serverid);
+
 	inet::service const service_;
 	// in memory representation of each file
 	std::unordered_map<std::string, file_info> files_;
 	// metadata for each chunk server
-	std::unordered_map<uint64_t, chunk_server_info> chunk_server_metadata_;
+	std::unordered_map<serverid, chunk_server_info> chunk_server_metadata_;
 	// map from chunkid to list of chunk servers
 	std::unordered_map<chunkid, std::vector<chunk_server_info*> > chunk_locations_;
 	// persistent/on disk copy of files_
--- a/include/sadfs/sadmd/sadmd.hpp
+++ b/include/sadfs/sadmd/sadmd.hpp
@@ -13,7 +13,7 @@ namespace sadfs {
 
 using chunkid = uint64_t;
 using serverid = uint64_t;
-using ticks = std::chrono::steady_clock::rep;
+using tick = std::chrono::steady_clock::rep;
 
 // all the information needed about a chunk server
 struct chunk_server_info
@@ -21,7 +21,7 @@ struct chunk_server_info
 	inet::service service;
 	uint64_t max_chunks;
 	uint64_t chunk_count;
-	ticks ttl;
+	tick ttl;
 };
 
 // all the information needed about a file
--- a/src/sadmd/sadmd.cpp
+++ b/src/sadmd/sadmd.cpp
@@ -115,18 +115,19 @@ process_message(sadfs::socket const& sock)
 namespace sadfs {
 
 namespace time{
-ticks
-in_1_min()
-{
-	using namespace std::literals;
 
-	return (std::chrono::steady_clock::now() + 1min).time_since_epoch().count();
+tick
+future_tick(std::chrono::minutes delta) noexcept
+{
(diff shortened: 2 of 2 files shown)


include/sadfs/sadmd/sadmd.hpp
```diff
@@ -12,14 +12,16 @@
 namespace sadfs {
 
 using chunkid = uint64_t;
+using serverid = uint64_t;
+using nanosec = std::chrono::steady_clock::duration::rep;
```
damanmkumar: `steady_clock` has a type `rep` as well, which I think is the same as `steady_clock::duration::rep`. Using `steady_clock::rep` is cleaner.

Also, it doesn't represent nanoseconds, but the number of ticks since the epoch. So, the alias should maybe be `ticks`?

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -54,11 +56,19 @@ class sadmd
 	// returns true if the database contains a file with the given name
 	bool db_contains(std::string const&) const;
 
+	// functions for maintaining chunk servers 
+
+	// returns true on success
+	bool add_server_to_network(serverid, char const*, int, uint64_t);
+	void remove_server_from_network(serverid) noexcept;
+	void register_server_heartbeat(serverid) noexcept;
+	bool is_active(serverid) const noexcept;
```
damanmkumar: This works for now, but should be handled by a different entity. `sadmd` now has way too many responsibilities.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -12,14 +12,16 @@
 namespace sadfs {
 
 using chunkid = uint64_t;
+using serverid = uint64_t;
```
damanmkumar, line 15: We should merge `chunkid` and `serverid` into `uuid`. Their "type" is essentially the same.

src/sadmd/sadmd.cpp
```diff
@@ -114,6 +114,22 @@ process_message(sadfs::socket const& sock)
...
 namespace sadfs {
 
+namespace time{
+nanosec
+in_1_min()
+{
+	using namespace std::literals;
+
+	return (std::chrono::steady_clock::now() + 1min).time_since_epoch().count();
+}
```
damanmkumar: Let's get rid of this. We need a function that accepts a time interval and returns the corresponding TTL.

damanmkumar: Maybe a template will be a good idea.

src/sadmd/sadmd.cpp
```diff
@@ -114,6 +114,22 @@ process_message(sadfs::socket const& sock)
...
+
+nanosec 
+current()
+{
+	return std::chrono::steady_clock::now().time_since_epoch().count();
+}
```
damanmkumar: This should be a wrapper around the function that accepts a time interval, with the interval set to `0min` or `0s`.

src/sadmd/sadmd.cpp
```diff
@@ -230,4 +246,45 @@ save_files() const noexcept
...
+			<< uuid
+			<< " which is already on the network\n";
+		return false;
+	}
+	chunk_server_metadata_.emplace(
+		uuid,
+		chunk_server_info{
+			inet::service(ip, port),
+			available_chunks,
+			0, // chunk count is zero to start
```
damanmkumar: When a chunk server registers with a master server upon restart, it should advertise the number of chunks it owns. So, it will be 0 only in special cases.

src/sadmd/sadmd.cpp
```diff
@@ -230,4 +246,45 @@ save_files() const noexcept
...
+void sadmd::
+remove_server_from_network(serverid id) noexcept
+{
+	chunk_server_metadata_.erase(id);
+}
+
+void sadmd::
+register_server_heartbeat(serverid id) noexcept
+{
+	auto server = &chunk_server_metadata_.at(id);
```
damanmkumar: We should check if the `id` is a registered server in the first place. `at(id)` will throw an exception if `id` does not exist, if I'm not mistaken.

damanmkumar: Why are we using a pointer here? A reference will work just fine, won't it?

src/sadmd/sadmd.cpp
```diff
@@ -230,4 +246,45 @@ save_files() const noexcept
...
+	server->ttl = time::in_1_min();
```
damanmkumar: This will call/use our parameterized function. Maybe: `server.ttl = time::in(1min);`

damanmkumar (changes requested):
A few comments

include/sadfs/sadmd/sadmd.hpp
abmono93: This is for readability. e.g. on lines 71-73 it makes the distinction between `chunk_server_metadata_` and `chunk_locations_` very intuitive. Also something like `is_active(serverid)` is self documenting vs. `is_active(uuid)`

src/sadmd/sadmd.cpp
abmono93: It also needs to tell the master which chunks it owns, not just how many, so that `chunk_locations_` is up to date. I think we'll need one function like `add_chunk_to_server()` which updates `chunk_locations_` and increments this field. After the chunk server has joined the network it can send individual control messages for each chunk, just like when it gets new chunks.

src/sadmd/sadmd.cpp
abmono93: I can't remember exactly but it had something to do with the fact that there's no default constructor for `inet::service` which is a member variable of `chunk_server_info`. Maybe when was when trying to use brackets. Anyway I've changed it so we just update ttl directly without creating a local variable

include/sadfs/sadmd/sadmd.hpp
abmono93: I agree but maybe just `tick`. My thinking is it's `int` not `ints` regardless of how many

damanmkumar: `tick ttl;` definitely doesn't read right, because `ttl` is not a tick. And now that we discussed it, neither does `ticks ttl;`, because `ttl` is not (a) "ticks."
The number of people in a room is a "head count."
The number of ticks after which an entry is no longer valid is a... "tick count?"
Like so: `tick_count ttl;`

BUT, what `ttl` really is is a point in time in the future, that we are representing as an integer that encodes the number of ticks since an epoch. So, `time_point` is probably the most representative of our idea. What do you think?

damanmkumar: With `time_point`, our functions will read better as well. `time::in(duration d)` returns a `time_point` which is time-duration `d` away from this moment in time.

include/sadfs/sadmd/sadmd.hpp
damanmkumar: Well, `map<uuid, data> server_metadata` and `map<uuid, server_info> chunk_locations` tell the reader that they are mapping the uuid of a server to its metadata, and the uuid of a chunk to its location. `is_active` is a very generic function name, and shouldn't be a private member function of the master server anyway. So, I'm not convinced with the second reason.

But, being explicit about `serverid` and `chunkid` in the declarations of the unordered maps is convincing enough.

src/sadmd/sadmd.cpp
damanmkumar: Sending a message over the network has a sizable overhead. So, we definitely should exploit this registration message to also send in the list of chunks that the chunk server owns.

However, the number of chunks on a server could theoretically be huge enough to make our protobuf larger than 1 MB (which is when google suggests we think of an alternate strategy). So, how about we send in the list of chunks with the registration message as well?

In the case that the list is very large, we send an empty list, and then register the chunks in batches. As a first step, implementing just the first case should be good enough. In that case, the initial chunk count should be a parameter (or however we're passing in `available_chunks`.

That reminds me -- `available_chunks` sounds like a list/count of chunks that are available for clients to read/write.

src/sadmd/sadmd.cpp
```diff
@@ -230,4 +247,60 @@ save_files() const noexcept
...
+{
+	if (!chunk_server_metadata_.count(id))
+	{
+		std::cerr << "Error: received heartbeat from server "
+			<< id
+			<< " which is not on the network\n";
+		return;
+	}
+	using namespace std::literals;
+	chunk_server_metadata_.at(id).ttl = time::in(1min);
```
damanmkumar: Can we wrap `time::in(1min)` into a function or make the `1min` a default value? Just so we don't introduce bugs later if we decide to change this to a different value, and not update it in some places.

damanmkumar (changes requested):
A couple more comments

include/sadfs/sadmd/sadmd.hpp
abmono93: That's kind of what I was going for with `nanosec` I guess. It's a particular nanosecond that could be in the future or past. But you're right that if a tick isn't a nanosecond it's not the right name. I can't think of something better than `time_point` so I'll go with that

src/sadmd/sadmd.cpp
abmono93: I came across a nice solution while going to change this: we just have an optional argument that defaults to one minute. It's not as intuitive to read as `time::in(1min)` so maybe `in()` should be renamed. But there's the added benefit that you don't need to have `namespace std::literals` every time you call this which is a little clunky

src/sadmd/sadmd.cpp
abmono93: Yea that is true. What about a separate control message for a server to "claim" one or more chunks in a batch? Then adding a server takes 1 message for a brand new server and 2 for a server restarting.

And I agree that `available_chunks` isn't a great name. I just didn't want to confuse it with `max_chunks` and this is the best I came up with. Do you have any ideas?

src/sadmd/sadmd.cpp
abmono93: After reviewing this again I came up with a slightly different approach. I renamed `chunk_server_info.tll` to `chunk_server_info.expiration_point` since it's a better description of how this variable is used. Then I defined 2 constants `server_ttl` and `file_ttl`, and `in()` is now called `future()` and is meant to take in one of those two constants to set the expiration point. lmk if you think this works

src/sadmd/sadmd.cpp
shu-g: Even if we assume that the chunk server holds 1TB data, we will have 1TM/64MB chunks i.e. 2^(40 - 26) or 2^14 chunks. If the UUID for each chunk consumes 8B, the space consumed by the list will be 2^(14+3) B which is less than 1MB.
So, I also think we should send the UUIDs with the control message at this point.
How about `stored_chunks`?

damanmkumar: UUIDs are 128-bits i.e., 16 bytes. Even if that is the case, Shubham's observation holds true. A protobuf will be a little over 256KB

src/sadmd/sadmd.cpp
```diff
@@ -114,6 +114,26 @@ process_message(sadfs::socket const& sock)
...
 namespace sadfs {
 
+namespace time{
+
+using namespace std::literals;
+constexpr auto server_ttl = 1min;
+constexpr auto file_ttl = 1min;
+
+time_point
+future(std::chrono::minutes delta) noexcept
```
damanmkumar: `future` has a different meaning in computer science: https://en.wikipedia.org/wiki/Futures_and_promises
It also doesn't read right. How about `time::from_now(server_ttl)/time::from_now(1min)` ?

src/sadmd/sadmd.cpp
```diff
@@ -114,6 +114,26 @@ process_message(sadfs::socket const& sock)
...
+{
+	return (std::chrono::steady_clock::now() + delta).time_since_epoch().count();
+}
+
+time_point 
+current() noexcept
```
damanmkumar: I feel `time::now()` sounds better, and is used everywhere in general. Agree?

src/sadmd/sadmd.cpp
```diff
@@ -114,6 +114,26 @@ process_message(sadfs::socket const& sock)
 
 namespace sadfs {
 
+namespace time{
+
+using namespace std::literals;
+constexpr auto server_ttl = 1min;
+constexpr auto file_ttl = 1min;
```
damanmkumar, line 121: This is a great idea!

damanmkumar (changes requested):
Two comments

damanmkumar (commented):
approved

PR #44, opened 2020-11-25
made minor optimizations

damanmkumar:
* clear `std::vector<uint64_t> chunkids` before deserializing to it, in case it is being reused
* pre-allocate storage for `chunkids` before deserializing ids to it
* removed redundant check for data in string buffer -- the `protobuf` will parse empty strings correctly

### DIFF
--- a/src/sadmd/sadmd.cpp
+++ b/src/sadmd/sadmd.cpp
@@ -57,13 +57,18 @@ void
 deserialize(std::vector<uint64_t>& chunkids, 
 	std::string const& existing_chunks)
 {
-	auto chunkid_pb = sadfs::proto::internal::chunkids{};
 	// parse string into protobuf object
+	auto chunkid_pb = sadfs::proto::internal::chunkids{};
 	chunkid_pb.ParseFromString(existing_chunks);
-	// copy protobuf object into vector
-	for (auto i = 0; i < chunkid_pb.chunkid_size(); i++)
+	
+	// clear chunkids, and allocate enough space for contents
+	chunkids.clear();
+	chunkids.reserve(chunkid_pb.chunkid_size());
+	
+	// copy items from protobuf object into vector
+	for (auto id : chunkid_pb.chunkid())
 	{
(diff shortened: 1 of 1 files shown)


src/sadmd/sadmd.cpp
```diff
@@ -57,13 +57,18 @@ void
...
 	chunkid_pb.ParseFromString(existing_chunks);
-	// copy protobuf object into vector
-	for (auto i = 0; i < chunkid_pb.chunkid_size(); i++)
+	
+	// clear chunkids, and allocate enough space for contents
+	chunkids.clear();
+	chunkids.reserve(chunkid_pb.chunkid_size());
+	
+	// copy items from protobuf object into vector
+	for (auto id : chunkid_pb.chunkid())
```
abmono93, line 74: Thanks this is much better

src/sadmd/sadmd.cpp
```diff
@@ -129,10 +134,7 @@ create_file(std::string const& filename, std::string const& existing_chunks)
 	if (!files_.count(filename))
 	{
 		auto info = file_info{};
-		if (existing_chunks.size() > 0)
-		{
-			deserialize(info.chunkids, existing_chunks);
-		}
+		deserialize(info.chunkids, existing_chunks);
```
abmono93, line 142: This is fine but the check was more to prevent `deserialize` from a bit of unnnecessary work.
i.e. in this updated version it will declare a protobuf object that never gets used, call parseFromString on the empty string, clear() the already empty vector created on line 136 and then see that the protobuf it created is still empty and do nothing. I figured it might save at least a few machine instructions to just put a check here

damanmkumar: That's super smart. How about moving this optimization into `deserialize()`? That way the code will be a lot cleaner at the call site, and we get this optimization every time without having to duplicate the check everywhere.

abmono93: Good idea

abmono93 (approved):

PR #46, opened 2020-11-29
partially (re-)implemented control messages

damanmkumar:
  to create control messages
- serializer classes manage serialization based on recipient
- socket class moved to `comm` namespace
- `inet` namespace renamed to `comm`
- `example/proto-msg` only serializes one message for now


src/proto/chunk.proto
```diff
@@ -0,0 +1,18 @@
...
+message chunk_request {
+	io_type type     = 1;
+	uint64  chunk_id = 3;
+}
+
+// generic request message
+message control_message {
+	sadfs.proto.host_id sender = 1;
+	oneof               msg {
+		chunk_request chunk_req = 11;
```
abmono93, line 16: What's the significance of 11?

src/proto/master.proto
```diff
@@ -0,0 +1,20 @@
...
+syntax = "proto3";
+package sadfs.proto.master;
+import "common.proto";
+
+// request for file IO
+message file_request {
+	io_type type     = 1;
+	string  filename = 2;
+	uint64  offset   = 3;
+	uint64  length   = 4;
```
abmono93, line 11: Is this meant to be the offset and length measured in bytes or chunks?

include/sadfs/msgs/messages.hpp
```diff
@@ -0,0 +1,106 @@
...
+	file_request,
+};
+
+// specifies a section of a file
+struct file_section
+{
+	std::string const filename;
+	std::size_t const offset;
+	std::size_t const length;
+};
```
abmono93, line 43: I guess that answers my other question about the protobuf definition.
I'm not against doing it this way but my understanding was that this logic would be taken care of on the client side and the servers would deal with chunks of data. i.e. it would be the client's responsibility to know what chunk it needs to access and send a request accordingly, e.g. it asks for the third chunk of file `main.cpp` and the master server will tell it the chunk's id and location. From what I can see this would be a simpler approach. We can talk about this elsewhere if it will be a long conversation

src/proto/chunk.proto
damanmkumar: These are IDs that protobufs use internally to identify the fields that are being sent/received over the network. Changing these IDs is apparently not a good idea for backward compatibility. So, I left the first 10 for other fields we might need to add later.

PR #48, opened 2020-11-30
renamed inet namespace to comm

damanmkumar:
NOTE: this will make the next pull-request easy to review and assess
- implemented the move assignment operator in the socket
  class
- fixed early termination bug in echod

### DIFF
--- a/src/comm/socket.cpp
+++ b/src/comm/socket.cpp
@@ -1,4 +1,4 @@
-/* implementation of sadfs::socket */
+/* implementation of sadfs::comm::socket */
 
 // sadfs-specific includes
 #include <sadfs/comm/socket.hpp>
@@ -9,7 +9,7 @@
 #include <sys/socket.h> // ::socket
 #include <unistd.h>     // ::close
 
-namespace sadfs {
+namespace sadfs { namespace comm {
 namespace {
 
 int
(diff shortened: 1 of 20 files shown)


src/comm/socket.cpp
```diff
@@ -71,6 +71,21 @@ socket(socket&& other) noexcept
...
+	domain_     = other.domain_;
+	type_       = other.type_;
+	descriptor_ = other.descriptor_;
+
+	// prevent other's destructor from closing the descriptor
+	other.descriptor_ = -1;
+
+	return *this;
+}
+
```
shu-g, line 88: We can't simply overwrite the members and need to `move` socket to `tmp` because we want the socket with previous values to be properly destroyed?
Can we call `move()` on the 3 members as well?

abmono93 (approved):

src/comm/socket.cpp
damanmkumar: Yes, you're right. We move `this` because we want it to be properly destroyed after the move assignment.

We could move the three members as well, but they're all PODs (plain old data types), and `std::move`s become copies. So, it would be extra code with no real difference.

shu-g (approved):
Looks good.

PR #51, opened 2020-12-01
(duplicate) partially (re-)implemented control messages

damanmkumar:
**NOTE:** duplicate of #46 since that is somewhat messed up, and shows unrelated changes
  to create control messages
- (de)serializer classes manage (de)serialization based on recipient
- serialization and deserialization of `file_request`s and `chunk_request`s supported


damanmkumar:
https://github.com/CS297P-SAD/sadfs/pull/46#discussion_r532825344

abmono93 (changes requested):
Only comment is still the change from dealing with bytes to dealing with chunks.
Other than that.. I don't totally understand how everything in here works. But I verified that it does work on my machine

damanmkumar:
`git rebase` gone wrong. This PR is beyond help.

PR #55, opened 2020-12-02
Create chunkid and uuid

### LINKED ISSUE #32: Master create chunkid
When a file needs another chunk to write to the master should create a new unique chunkid and assign it to that file

abmono93:
- Includes merged in uuid branch
- Redefined chunkid as alias for uuid instead of unsigned int
- Redefined internal protobuf for new chunkid definition, including wrapper class (which also replaces serialize and deserialize functions in sadmd.cpp)
- Implemented sadmd::add_chunk_to_file()
- Cleaned up error handling in create_file (I was planning to do this separately but it made sense to do it here during testing)

Closes #54 
Closes #32


src/comm/internal.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 package sadfs.proto.internal;
 
-message chunkid_container {
-    repeated uint64 chunkids = 1;
-}
\ No newline at end of file
+message file_chunks {
+    repeated bytes uuids = 1;
+}
```
damanmkumar: The file looks good, but it should be in `src/proto/internal/` so that it matches the `package` statement. Maybe we should rename the file to util(s).proto ?

include/sadfs/sadmd/internal.hpp
```diff
@@ -0,0 +1,59 @@
...
+#define SADFS_SADMD_INTERNAL_HPP
+
+// sadfs-specific includes
+#include <sadfs/proto/internal.pb.h>
+#include <sadfs/uuid.hpp>
+
+// standard includes
+#include <string>
+
+namespace sadfs {
```
damanmkumar: How about renaming this file to `util.hpp`, and moving its contents into the `sadfs::sadmd::util` namespace?

include/sadfs/sadmd/internal.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+// wrapper class around sadfs protobuf object for storing list of uuids
+class file_chunks
+{
+public:
+    file_chunks() = default;
+
+    void add_chunk(uuid const& id)
+    {
+        protobuf_.add_uuids(to_string(id));
```
damanmkumar: The `to_string` function returns a human-readable representation of the `uuid`, like `d0f34638-46c5-4c03-8ef7-7f0004af99fd`. We should be using the `uuid::serialize` function to serialize the `uuid` into 8 bytes.

One way to do it would be to create a temporary string, serialize the `uuid` to it, and move it into the protobuf (to avoid a copy).

Another way would be to use the `string* add_foo()` method ([reference](https://developers.google.com/protocol-buffers/docs/reference/cpp-generated#repeatedstring)) to get a pointer to a newly added string, and serialize the `uuid` to it by dereferencing the pointer.

`example/uuid.cpp` shows how to use a `std::back_inserter` to serialize a `uuid` into a `std::string`.

include/sadfs/sadmd/internal.hpp
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    std::string 
+    operator[](unsigned int const i)
+    {
+        //TODO: handle error.. or convert this to an iterator?
+        if (i >= protobuf_.uuids_size()) 
+        {
+            return std::string{};
+        }
```
damanmkumar: Let's not handle what is obviously a programming error. This will help hide bugs in our code. We should either directly call `protobuf_.uuids(i)`, or throw an exception after performing bounds checking.

That said, what is the use case for this operator? At the server side, we simply serialize from a `vector<uuid>`, and at the client side deserialize into a `vector<uuid>`, right?

In which case, we should maybe add a constructor that takes a `vector<uuid>`, and a `deserialize(std::vector<uuid>&)`function?

include/sadfs/sadmd/internal.hpp
```diff
@@ -0,0 +1,59 @@
...
+        return protobuf_.uuids(i);
+    }
+
+    size_t size()
```
damanmkumar: Let's specify that this is `std::size_t` so that there is no ambiguity. It's fine to not specify `std::` for `uint64_t` since that specifies the width explicitly.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -1,18 +1,26 @@
...
 #define SADFS_SADMD_SADMD_HPP
+
+// sadfs specific includes
 #include <sadfs/comm/inet.hpp>
 #include <sadfs/comm/socket.hpp>
+#include <sadfs/uuid.hpp>
+#include "internal.hpp" // file_chunks object
+
+// extrenal includes
+#include <boost/container_hash/hash.hpp>
```
damanmkumar: No longer required, right?

include/sadfs/sadmd/internal.hpp
abmono93: Good point, we don't really need error checking.

The intention for this class was to use an internal protobuf instead of a vector. The use case is when a client asks for the ith chunk of `filename`, we can find it in `files[filename].chunkids[i]` as if it were in a vector.

damanmkumar: Oh, damn, I had it completely mixed up in my head. Yup, this looks good. If you decide to throw, rename the function to `at(std::size_t i)`. If you decide to simply call `uuids(i)`, the operator overload is perfect.

damanmkumar: One comment, though. Since this is an in-line definition, let's not move the name of the function to a separate line.

src/sadmd/sadmd.cpp
```diff
@@ -159,14 +120,16 @@ create_file(std::string const& filename, std::string const& existing_chunks)
 	if (!files_.count(filename))
 	{
 		auto info = file_info{};
-		deserialize(info.chunkids, existing_chunks);
+		info.chunkids.deserialize(existing_chunks);
```
damanmkumar: I don't have a comment for this line, but the function name `create_file` doesn't make sense in the context of `load_files`, right? We have a different use case for file "creation." Maybe `load_file` is better?

include/sadfs/sadmd/internal.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+namespace sadfs {
+
+// wrapper class around sadfs protobuf object for storing list of uuids
+class file_chunks
+{
+public:
+    file_chunks() = default;
+
+    void add_chunk(uuid const& id)
```
damanmkumar: Since we're `using chunkid = sadfs::uuid`, let's keep it consistent, and use the same type alias here as well.

src/comm/internal.proto
```diff
@@ -1,6 +1,6 @@
 syntax = "proto3";
 package sadfs.proto.internal;
 
-message chunkid_container {
-    repeated uint64 chunkids = 1;
-}
\ No newline at end of file
+message file_chunks {
+    repeated bytes uuids = 1;
```
damanmkumar: For consistency, how about we name this field `chunkids`?

Also, ignore my comment about moving this file to `src/proto/`. That change hasn't been merged in yet.

src/sadmd/sadmd.cpp
```diff
@@ -294,10 +257,25 @@ is_active(serverid id) const noexcept
...
+add_chunk_to_file(std::string const& filename)
+{
+	if (!files_.count(filename))
+	{
+		std::cerr << "Error: cannot add new chunk to file "
+				  << filename
+				  << ": file does not exist\n";
+		return;
+	}
+	auto new_chunkid = chunkid::generate();
```
damanmkumar, line 278: The master server will generate a chunkid and send it to the client upon a request to create a new chunk. Then, upon successful write to the main chunk server, and receipt of a control message from the chunk server informing the master server of a new chunk, the master will add that chunkid to its in-memory store. So, this method should take a chunkid as a parameter.

damanmkumar (changes requested):
Left a few comments, all of them minor

src/sadmd/sadmd.cpp
abmono93: This function can be used both to create new files in an active system and to "load" files on a master that's just started. The way I think of it is we have recorded there was a file with such and such properties. Now that the system is restarting, we need to create (or I guess recreate) it.

That is also the reason we have the existing chunks as an optional argument.

src/sadmd/sadmd.cpp
abmono93: I don't think that's the right approach. The use case here is a client wants to add a new chunk to a file, either because the file has been created and nothing written to it yet or the last chunk of the file is full. So the master needs to generate a new unique ID for this new chunk and should record this chunk exists upon creation for a couple reasons:
- one of the inherent properties of this chunk is its the first (or second etc..) chunk of `filename`, so this ensures that the list of chunks is in the right order
- now when a server tells the master it has written this chunk, the master is already aware it exists and adds that server to its (previously empty) list of locations. This ensures no chunks are added to the system except those generated by the master

abmono93: Actually ignore this^ I have a better idea

src/sadmd/sadmd.cpp
damanmkumar: It doesn't make sense for the two use cases to use the same function. At the very least, this function should be called `load_file`, with `create_file` maybe being a wrapper around it like
```
create_file(... filename)
{
    ...
    load_file(filename, std::string{}); // new file -- no chunks
    ...
}
```
simply because we are implementing file creation by loading it now, and assuming it will be persisted later.

src/sadmd/sadmd.cpp
abmono93: Ok my idea didn't end up working out but I renamed this function to `append_chunk_to_file` and changed some other things. Lmk if the current version makes sense to you.

include/sadfs/sadmd/internal.hpp
abmono93: I couldn't easily get this to work without mangling some of the uuids after restarting the server... obviously I'm doing something wrong. But this protobuf isn't getting sent over the network anyway, it's just for sadmd to use instead of a vector and to speed up saving files in the database. So I think it's okay the way it is even if it's not ideal

include/sadfs/sadmd/util.hpp
```diff
@@ -0,0 +1,55 @@
...
+        protobuf_.SerializeToString(&pb_str);
+        return pb_str;
+    }
+
+    void deserialize(std::string const& str)
+    {
+        protobuf_.ParseFromString(str);
+    }
+
+    chunkid operator[](unsigned int const i)
```
damanmkumar, line 38: `protobuf_.chunkids(i)` expects an `int`. Let's change `i`to just be an `int`. This will be an issue once we enable warnings on our code base, since it will complain about narrowing.

include/sadfs/sadmd/util.hpp
```diff
@@ -0,0 +1,55 @@
...
+#define SADFS_SADMD_UTIL_HPP
+
+// sadfs-specific includes
+#include <sadfs/proto/internal.pb.h>
+#include <sadfs/uuid.hpp>
+
+// standard includes
+#include <string>
+
+namespace sadfs { namespace util {
```
damanmkumar, line 11: Namespaces should (mostly) match the location they're in or logical partitions. So, this should be in `sadfs::sadmd::util` because it contains utility components for `sadmd` and all of sadfs in general.

include/sadfs/sadmd/util.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+// wrapper class around sadfs protobuf object for storing list of uuids
+class file_chunks
+{
+public:
+    file_chunks() = default;
+
+    void add_chunk(chunkid const& id)
+    {
+        protobuf_.add_chunkids(to_string(id));
```
damanmkumar, line 23: Let's address this during our meeting tomorrow morning.

src/sadmd/sadmd.cpp
```diff
@@ -153,22 +114,25 @@ start()
...
+{
+	create_file(filename);
+	auto file_chunkids  = &(files_[filename].chunkids);
+	file_chunkids->deserialize(existing_chunks);
+	reintroduce_chunks_to_network(*file_chunkids);
+}
+
 void sadmd::
-create_file(std::string const& filename, std::string const& existing_chunks)
+create_file(std::string const& filename)
```
damanmkumar, line 126: Shouldn't this be the other way around, where `create_file` calls `load_file` with an empty `std::string`? Because, logically, when we want to create a file, we "load" it into our in-memory data structure -- assuming that it will be persisted to disk during shutdown, thus creating it.

src/sadmd/sadmd.cpp
```diff
@@ -153,22 +114,25 @@ start()
 	}
 }
 
+void sadmd::load_file(std::string const& filename, std::string const& existing_chunks)
+{
+	create_file(filename);
+	auto file_chunkids  = &(files_[filename].chunkids);
```
damanmkumar, line 120: We don't have to play with pointers here. References will work without creating copies, like it currently does with the call to `reintroduce_chunks_to_network`.
```
auto& file_chunkids = files_[filename].chunkids;
file_chunkids.deserialize(...);
reintroduce...(file_chunkids);
```

```
void reintroduce_chunks_to_network(file_chunks& chunks);
```

src/sadmd/sadmd.cpp
```diff
@@ -153,22 +114,25 @@ start()
 	}
 }
 
+void sadmd::load_file(std::string const& filename, std::string const& existing_chunks)
```
damanmkumar, line 117: The function name needs to appear on a new line.

src/sadmd/sadmd.cpp
damanmkumar: Also, if we add a constructor to file_chunks that takes a string and deserializes the protobuf from it, our code will be much cleaner. We will no longer need to add an entry to `files_` and then retrieve the value to  deserialize it from a string.

src/sadmd/sadmd.cpp
damanmkumar: The scenario I am concerned about is this:
- master server generates a chunkid and adds it to the required file, and into `chunk_locations_`
- the write operation fails

Now we have an inconsistent state. You raised a terrific point about adding a chunkid that we cannot guarantee was generated by us. One way to address both issues is to add this newly generated chunkid into a `tentative_chunkids` list, and leaving the `files_` structure unmodified.

Later, when chunk servers inform the master server of new chunkids, we can verify it against this list before modifying `files_` and `chunk_locations_`.

include/sadfs/sadmd/internal.hpp
damanmkumar: Can you take a look at #57 ? If it doesn't work out of the box with your tests, we can address it sometime later.

abmono93: Thanks, that worked

src/sadmd/sadmd.cpp
abmono93: That's a good point but I think we can handle that without creating another data structure that would need to be managed. 
- The write operation fails for some reason and the client gives up
- Eventually their write lock on this file expires
- Now the next time someone tries to write, master behaves exactly as in any other situation when the last chunk of the file isn't full.

 i.e. instead of generating another chunkid, it looks at this file and sees that more data can fit on the last chunk. It then tells the client the chunkid and which server(s) to write it to. The only potential issue is the master has to recognize what to do when `server_locations_[chunkid]` is empty. But that also happens any time new data is written to a chunk, so we already have to implement this decision process.

include/sadfs/sadmd/util.hpp
abmono93: I just tried implementing this but realized`sadmd` is already the name of a class

damanmkumar (approved):
LGTM

PR #57, opened 2020-12-04
modified uuid IO to/from file_chunks

damanmkumar:
`file_chunks` now uses lesser memory


abmono93 (approved):
Passed the tests I've been using which do the following:
- append chunks to files
- prints all the files in the system and the chunkids associated with them
- after saving and restarting sadmd I verified the same chunkids were associated with the same files

PR #60, opened 2020-12-04
(re)implementation of control messages

damanmkumar:
  to create control messages
- (de)serializer classes manage (de)serialization based on recipient
- serialization and deserialization of `chunk_location_request`s
  and `chunk_request`s supported

essentially a clean copy of #51


include/sadfs/msgs/chunk_messages.hpp
```diff
@@ -0,0 +1,57 @@
...
+private:
+	proto::chunk::chunk_request protobuf_{};
+
+	// provide embed/extract functions access to private members
+	friend bool embed(chunk_request const&, message_container&);
+	friend bool extract(chunk_request&, message_container const&);
+};
+
+// declarations
+bool embed(chunk_request const&, message_container&);
```
abmono93, line 41: Yea message_container is good. It makes something like this fairly intuitive.

example/proto-msg.cpp
```diff
@@ -1,139 +1,99 @@
...
+main(int argc, char** argv)
 {
-	auto echoed_req = comm::msgs::chunk_request{};
-	if (!echoed_req.recv(in))
+	auto fr = msgs::master::chunk_location_request
 	{
-		std::cerr << "error: could not receive chunk request\n";
-		std::exit(1);
-	}
+		msgs::io_type::read,
```
abmono93, line 37: Just a note - if we end up going with the file lock, this will always be read. The client wouldn't request a chunk location in the context of a write. So this may end up being superfluous.

abmono93 (approved):
Left some general comments but didn't see anything that requires changes

example/proto-msg.cpp
damanmkumar: Yeah, that makes sense!

PR #64, opened 2020-12-06
Master serves chunk location request

abmono93:
- Introduced sadmd::process() for chunk_location_request type messages. Should be done for each type of message a master can receive
- Implemented add_chunk_to_server() for testing purposes, but we will need it for actual behavior soon
- Implemented choose_best_server() which simply chooses the first server on the list (should be changed to some algorithmic decision)
- Created example/send-req to demonstrate
- Replaced sadmd::process_message() with an infinite loop that assumes an incoming message is a chunk_location_request. This will need to be expanded to determine what kind of request is coming in first


src/sadmd/sadmd.cpp
```diff
@@ -289,4 +275,113 @@ reintroduce_chunks_to_network(util::file_chunks ids)
...
+void sadmd::
+process(msgs::channel& ch, msgs::master::chunk_location_request& clr)
+{
+	auto ok = true;
+	auto servers = std::pair<chunk_server_info*, std::string>{};
+	chunkid id;
+
+	if ((ok = is_valid_chunk(clr.filename(), clr.chunk_number())))
+	{
+		id = files_[clr.filename()].chunkids[clr.chunk_number()];
```
damanmkumar: This is the third time a lookup happens for the filename in `files_`. We can limit it to just one by using the `find` member function, which returns an iterator that can be compared with `files_.end()` to find out if the lookup succeeded. The advantage is that we don't to perform lookups again, when we want to do something with the value in the map, like verifying chunk number validity or looking up the UUID of the chunk.

The signature of `is_valid_chunk` will need to change to accept a std::unordered_map<..., ...>::iterator, but that function can simply become a local lambda function since it only does very basic validation.
```
auto validate = [](auto it, auto chunk_number)
{
    if (chunk_number >= it->chunkids.size()) // safe since we don't validate if it is invalid
        ...
};

auto it = files_.find(filename);
if (it == files_.end()) // key does not exist, handle it
ok = validate(it, clr.chunk_number());
...
```

src/sadmd/sadmd.cpp
```diff
@@ -289,4 +275,113 @@ reintroduce_chunks_to_network(util::file_chunks ids)
...
+	auto response = msgs::client::chunk_location_response
+	{
+		ok,
+		ok ? servers.first->service : comm::service{"0.0.0.0", 0},
+		id, // junk if !ok
+		servers.second
+	};
+	msgs::client::serializer{}.serialize(response, ch);
+	ch.flush();
+	ch.close();
```
damanmkumar: We shouldn't be calling `close()` on the channel since that will permanently close the channel, and further requests from the client will be lost.

src/sadmd/sadmd.cpp
```diff
@@ -289,4 +275,113 @@ reintroduce_chunks_to_network(util::file_chunks ids)
...
+				  << " of "
+				  << filename
+				  << '\n';
+		return false;
+	}
+	return true;
+}
+
+std::pair<chunk_server_info*, std::string> sadmd::
+choose_best_server(std::vector<chunk_server_info*>& servers, bool include_rest)
```
damanmkumar: How about `choose_servers` that returns three (or another compile-time constant) `chunk_server_info` pointers? The name now doesn't tell us that it is doing more than just choosing one server to host a chunk.

src/sadmd/sadmd.cpp
```diff
@@ -289,4 +275,113 @@ reintroduce_chunks_to_network(util::file_chunks ids)
...
+		std::cerr << "Error: list of server locations empty\n";
+		return {nullptr, {}};
+	}
+	//TODO: make decision more complicated....
+	auto best = servers[0];
+	if (include_rest)
+	{
+		return {best, all_servers_except(servers, best)};
+	}
+	return {best, {}};
```
damanmkumar: Until we finalize a strategy to choose 3 servers, how about just returning `servers[0...2]`?

damanmkumar: And by then, we can add a protobuf to use as our secret payload. It'll simple be way easier to add/remove stuff to it

src/sadmd/sadmd.cpp
```diff
@@ -289,4 +275,113 @@ reintroduce_chunks_to_network(util::file_chunks ids)
...
+add_chunk_to_server(chunkid cid, serverid sid)
+{
+	if (!is_active(sid)) return;	
+	auto server_ptr = &(chunk_server_metadata_.at(sid));
+
+	if (server_ptr->chunk_count >= server_ptr->max_chunks)
+	{
+		std::cerr << "Error: chunk server full\n";
+		return;
+	}
```
damanmkumar: We will be adding a chunk to a server only after the sever itself has informed that a chunk has been stored on it, right? So, this check will never be required. Also, how about asking the chunk server to simply send us the new `chunk_count` instead of us keeping count separately? That way, there will be no possibility of the two counts diverging unless an error has occurred.

src/sadmd/sadmd.cpp
```diff
@@ -289,4 +275,113 @@ reintroduce_chunks_to_network(util::file_chunks ids)
...
+						  + ':'
+						  + std::to_string(to_int(server_ptr->service.port()))
+						  + ';';
+		}
+	}
+	return other_servers;
+}
+
+void sadmd::
+add_chunk_to_server(chunkid cid, serverid sid)
```
damanmkumar, line 351: Since there is a possibility of failure to add a chunk to a server, should this function return a `bool`?

include/sadfs/msgs/client_messages.hpp
```diff
@@ -21,13 +21,14 @@ class chunk_location_response
 {
 public:
 	chunk_location_response() = default;
-	chunk_location_response(bool ok, comm::service const& service,
-	                        chunkid chunk_id, std::string const& payload);
+	chunk_location_response(bool ok, chunkid chunk_id);
+	void add_service(comm::service const&);
+	void set_ok(bool);
```
damanmkumar: I see that `add_valid_servers` conditionally sets `ok` to `false` if there aren't any servers available. That pattern isn't great because the function has side effects that aren't immediately apparent. It also breaks the single responsibility principle. How about moving that decision to a different point in time, when we can determine authoritatively if `ok` should be `true` or `false`?

With a change in name, `add_valid_servers` can return a `std::vector<chunk_server_info*>` (or better yet, a `std::vector<comm::service>`), which will then let one function (`process()`) determine the value of `ok`.

The constructor of `chunk_location_response` can then simply be:
```
chunk_location_response(bool ok, std::vector<comm::service> const&, chunkid)
{
    ... set ok ...
    if(ok)
    {
        ... set server_ips and ports ...
    }
}
```

include/sadfs/msgs/client_messages.hpp
```diff
@@ -66,6 +61,12 @@ chunk_id() const noexcept
 	return id;
 }
 
+inline int chunk_location_response::
+num_locations() const
+{
+	return protobuf_.server_ips_size();
```
damanmkumar: I expected `server_ips_size()` to return a `std::size_t`, I'm surprised!

include/sadfs/msgs/client_messages.hpp
abmono93: I agree this shouldn't have a side effect. But I wanted to avoid looping through the servers twice, which we have to do if we put it in an in-memory intermediate container like a vector. So I think instead we should move the check from the end of `add_valid_servers` to process/handle, and keep `add_valid_servers` the way it is so it adds directly to the protobuf field avoiding the second loop.

abmono93:
This can't be meged yet. I need to rebase because of changes made in the last PR

abmono93:
> This can't be meged yet. I need to rebase because of changes made in the last PR

Now it can be merged

damanmkumar (approved):
made a few formatting changes and botched the rename of `chunk_location_request::num_locations()` to  `chunk_location_request::locations_size()`. I now realize that we should rename `service(int)` to `services(int)` or `servers(int)` and this function to `{services|servers}_size` accordingly (for consistency with the protobuf code). I'll do it in another PR. I'm merging this in now.

PR #65, opened 2020-12-06
implemented a message processor class

damanmkumar:
commit b6e35ac50f9d69eb9bc8c62856a65f2c1b11e923
Author: Daman Mohan Kumar <kumardm@uci.edu>
Date:   Sun Dec 6 14:14:36 2020 -0800

    implemented a message processor class

    - currently master server only
    - requires handler class to implement functions named
      handle for every control message type it wants to handle
    - separated messages, serializers, and deserializers into
      master/, chunk/, and client/ directories
    - added dummy-sadmd example program that responds to
      chunk_location_requests with a freshly generated uuid
    - modified proto-msg to only send requests
    - fixed message container name in client.proto and relevant
      wrapper classes


example/dummy-sadmd.cpp
```diff
@@ -0,0 +1,71 @@
...
+		while (true)
+		{
+			info("waiting for connections...");
+			auto sock = listener.accept();
+			info("accepted connection");
+			serve_requests(std::move(sock));
+		}
+	}
+
+	void serve_requests(msgs::channel ch)
```
abmono93, line 27: Is there any reason we can't put this in actual `sadmd.cpp`?

abmono93 (approved):

example/dummy-sadmd.cpp
damanmkumar: None at all. I simply used this to test and make sure everything worked as expected, while I was still ironing things out.

PR #72, opened 2020-12-08
Renamed chunk_locations to chunk_metadata_

### LINKED ISSUE #70: sadmd.chunk_locations_ should be renamed and support versions
`sadmd.chunk_locations_`  should map from chunkids to the most recent version number for that chunk, and each server in the list of locations should be associated with a version number as well. With this added information the data structure would be more than just chunk locations and should therefore be renamed to `chunk_metadata_`.

Once this is set up, the` add_valid_servers()` function should be modified to only add servers containing the latest version in the context of a read.

abmono93:
- replaced chunk_locations with chunk_metadata, which stores locations as well as version information

- added field for latest version to chunk_location_request protobuf (to be forwarded to chunk server on a write)

- added check for latest version to valid_servers() when latest_only == true (to be used for reading)

closes #70


src/sadmd/sadmd.cpp
```diff
@@ -64,18 +64,30 @@ open_db()
...
+	else
+	{	
+		for (auto location : info.locations)
+		{
+			if (location.first->valid_until > time::now())
+			{
+				services.push_back(location.first->service);
+			}
 		}
 	}
```
damanmkumar: Minor suggestion. Can we use a lambda to filter chunk servers to reduce the nesting we have here? Something along the lines of:
```
auto is_active = [](auto location){ return ... > time::now(); };
auto is_active_latest_version = [](auto location){ return is_active(location) && ...; };
auto const& filter = latest_only ? is_valid_latest_version : is_valid;
for (auto location : ...)
{
    if (filter(location)) { ... push back ... }
}

File: src/sadmd/sadmd.cpp [hunk 2]
```diff
@@ -326,8 +339,10 @@ process(msgs::channel& ch, msgs::master::chunk_location_request& clr)
 	else if (validate(it, clr.chunk_number()))
 	{
 		id = it->second.chunkids[clr.chunk_number()];
-		servers = valid_servers(chunk_locations_[id], 
-					clr.io_type() == sadfs::msgs::io_type::read); 
+		auto& chunk = chunk_metadata_[id];
+		servers = valid_servers(chunk, 
+								clr.io_type() == sadfs::msgs::io_type::read);
```
damanmkumar: Can we add an inline comment here preceding the comparison to tell the reader what we're actually passing to `valid_servers`? Like `/*latest_only=*/clr.io_type() == ...`

include/sadfs/msgs/client/messages.hpp
```diff
@@ -23,11 +24,12 @@ class chunk_location_response
 public:
 	chunk_location_response() = default;
 	chunk_location_response(bool ok, std::vector<comm::service> const& services, 
-	                        chunkid chunk_id);
+				chunkid chunk_id, version version_num);
```
damanmkumar, line 27: Hey, quick question. Why does the client need to know which version of a chunk it is receiving? I can't think of a reason chunk servers need to know the version either.

damanmkumar (changes requested):
One question, one suggestion. Looks good otherwise

include/sadfs/msgs/client/messages.hpp
abmono93: The client doesn't need to know per se, but they need to forward this to chunk servers when writing because they needs to supply a version number when they update the master. Maybe that won't be necessary when we have a robust system for relinquishing a lock which makes it impossible for an update from a chunk server to be referring to older versions.

abmono93: Also we can't have the chunk servers track this internally in case they miss a version for some reason, maybe a restart. e.g if it has version 5 and gets a write for version 7, it needs to know to tell the master it has version 7

PR #73, opened 2020-12-08
implemented heartbeat machinery

damanmkumar:
- added `heart` class and `example/dummy-sadcd.cpp` that simply sends heartbeats for 5 seconds.
- implementing the sending of heartbeats and a corresponding response will be implemented in a following PR


src/sadcd/heart.cpp
```diff
@@ -0,0 +1,65 @@
...
+beat() const
+{
+	auto msg          = msgs::master::chunk_server_heartbeat{};
+	auto serializer   = msgs::master::serializer{};
+
+	// send heartbeats until requested to stop
+	while (true)
+	{
+		using namespace std::chrono_literals;
+		auto status = stop_token_.wait_for(1s);
```
abmono93, line 49: I don't know what best practice is for a system like this but I think every second might be overkill. If each heartbeat keeps the server valid on the master's side for a minute, I'd think every 15-30 seconds would be fine.

src/sadcd/heart.cpp
```diff
@@ -0,0 +1,65 @@
...
+start()
+{
+	if (stop_token_.valid())
+	{
+		throw sadfs::logic_error{"heart beating already"};
+	}
+
+	// start heartbeat
+	stop_token_ = stop_request_.get_future();
+	heartbeat_ = std::thread{[this](){ this->beat(); }};
```
abmono93, line 21: Wow I didn't know about `std::thread`. I can't believe how simple this is!

abmono93 (approved):
One comment that's more of a suggestion/question but otherwise looks good. Works on my end and the interface is clean.

PR #75, opened 2020-12-09
implemented chunk server heartbeats

### LINKED ISSUE #67: Implement heartbeat
The chunk servers should send a heartbeat message at regular intervals and the master should accept these messages and update the server's valid_until field

damanmkumar:
- master server does not respond yet
- added .clang-format
- to format, run command: `INPLACE=yes ./format.sh [FILE...]`
- sadcd exits gracefully when the master server isn't reachable
- lots of changes are simply formatting changes


abmono93 (commented):
I don't think this should close the issue since the master still needs to handle heartbeats. Other than that it looks good.

The master can handle the heartbeats easily by calling register_server_heartbeat(). But that requires the server to proviede it's uuid. I am working on that as part of the join request implementation which should be done soon. Which reminds me that the heartbeat messages shouldn't be empty, they need to contain sadcd's serverid...

damanmkumar:
> I don't think this should close the issue since the master still needs to handle heartbeats. Other than that it looks good.
> 
> The master can handle the heartbeats easily by calling register_server_heartbeat(). But that requires the server to proviede it's uuid. I am working on that as part of the join request implementation which should be done soon. Which reminds me that the heartbeat messages shouldn't be empty, they need to contain sadcd's serverid...

Oh, yeah, it shouldn't. PRs don't close our issues anyway, since we never merge into master.
Also, I don't know if I commented on your PR yet, but the `message_container` carries the sender's ID. So, we need not send the UUID as part of the control message itself. We haven't been sending it yet, though. Let's go forward with the current implementation, I'll change that up in a couple of hours.

shu-g (commented):
The changes look good to me.

damanmkumar:
this is a lost cause

PR #77, opened 2020-12-09
Sadcd joins network on startup

### LINKED ISSUE #68: Implement join request
The chunk servers should be able to send a request to the master to join the network and the master should be able to accept these messages and respond appropriately.

This includes chunk server generating a uuid for itself (and saving it for restart) and master adding it to chunk_server_metadata_

abmono93:
- sadcd now takes thee additional arguments - `master_ipaddress`, `master_port`, and `server_id`
- on startup sadcd sends a request to join the provided master ip and port

the branch is called daemonS because I thought I'd have the master respond with reject/accept. I don't know if that's totally necessary though depending on how hearbeats etc. work, so we can implement that later if it turns out we need to.

closes #68


src/sadcd/sadcd.cpp
```diff
@@ -13,15 +16,26 @@
 
 namespace sadfs {
 
+namespace constants {
+
+constexpr auto default_max_chunks = 1000;
```
damanmkumar: This will be much less verbose:
```
namespace defaults {
constexpr auto max_chunks = 1000;
}
```

src/sadcd/sadcd.cpp
```diff
@@ -13,15 +16,26 @@
...
 sadcd::
-sadcd(char const* ip, int port) : service_(ip, port)
+sadcd(char const* ip, int port, 
+      char const* master_ip, int master_port, 
+	  char const* server_id) : 
+	  service_(ip, port) , 
+	  master_(master_ip, master_port)
 {
-	// do nothing
+	serverid_ = serverid::from_string(server_id);
```
damanmkumar: Any particular reason we are not using the constructor initializer list like so:
```
...
master_(...),
serverid_(serverid::from_string(...))
```

src/sadcd/sadcd.cpp
```diff
@@ -64,4 +78,35 @@ process_message(comm::socket const& sock)
 	return result;
 }
 
+void sadcd::
+join_network()
```
damanmkumar: How about returning a `bool` indicating success/failure, which will then be used to decide what needs to be done? We can exit gracefully as opposed to via `std::exit`, which now feels like a very bad thing since destructors are not called when we exit.

src/sadcd/sadcd.cpp
```diff
@@ -64,4 +78,35 @@ process_message(comm::socket const& sock)
...
+			return msgs::channel{master_.connect()};
+		}
+		catch (std::system_error ex)
+		{
+			std::cerr << "[error]: failed to connect to master server\n"
+						<< ex.what() << "\n";
+			std::exit(1);
+		}
+	};
+	auto ch = establish_conn();
```
damanmkumar: Let's replace this by something like:
```
auto ch = channel{socket{domain::inet, type::stream, -1}}; // dummy channel
try
{
    ch = ... // get actual channel
}
catch
{
    // log error and return false
}
```

Let me quickly check if `channel` let's us initialize it with a dummy socket that way

src/sadcd/sadcd.cpp
```diff
@@ -64,4 +78,35 @@ process_message(comm::socket const& sock)
...
+	{
+		serverid_,
+		service_,
+		constants::default_max_chunks,
+		/*chunk_count=*/ 0
+	};
+
+	// send join_network_request
+	msgs::master::serializer{}.serialize(jr, ch);
+	ch.flush();
```
damanmkumar: Flush is unnecessary here, and we'll be looking for a positive response from the master here in the future, correct?

src/sadcd/sadcd.cpp
damanmkumar: Ok, channel let's you do it, but I need to fix it so that `channel::is_open` returns false in that case. I'll do that in a separate PR.

src/sadcd/sadcd.cpp
abmono93: Nope. That's better to because it will allow me to declare serverid_ as const now

src/sadcd/sadcd.cpp
abmono93: Most likely, yes

src/sadcd/sadcd.cpp
abmono93: Ok I marked this as TODO then. The current version just assumes the master server is running and the channel opens. When you modify `channel::is_open` you can make the change to support the error case. Does that work?

damanmkumar (approved):
LGTM

PR #79, opened 2020-12-10
Chunk write notify

### LINKED ISSUE #71: Implement notify master of chunk write
Chunk servers should be able to send a message notifying the master that it performed a write. It should provide the chunk id and version number. The master should be able to receive these messages and update it's internal info to reflect the change

abmono93:
- Created protobuf and wrapper class for the chunk write notification messages
- Implemented another `handle` on sadmd to update data structures accordingly
- Turns out storing the chunk locations as a vector of `<pointer, version>` pairs wasn't feasible (would have required looping through a list after every write notificatoin) so it is now a map from server ids to version numbers
- As a consequence I had to rewrite some other code in `sadmd.cpp` so it used chunk_metadata_ appropriately
- `add_chunk_to_server` returns a bool now
- Created a function in sadcd to notify the master of a successful write

Closes #71


example/send-req.cpp
```diff
@@ -137,15 +137,45 @@ join_network()
...
 
+void
+notify_chunk(serverid sid, chunkid cid, version v)
+{
+	auto cwn = msgs::master::chunk_write_notification
+	{
+		sid,
+		cid,
+		v,
+		"/mnt/a/file.dat"
```
shu-g: I don't think chunks have knowledge of which file they belong to on the chunk server end, the master can update `chunk_metadata_` and `chunk_server_metadata` without the filename.

shu-g (changes requested):
Left comment highlighting a potential issue, otherwise looks good.

example/send-req.cpp
abmono93: I agree this isn't ideal but because of other implementation choices we've made it's necessary for the time being. 
When the chunk is first created, we decided not to immediately add it to the files' metadata. Which means when the master receives a notification about a new chunk it won't know which file to associate the chunk with. The chunk server won't have to actually keep track of filenames, it can just forward it along from the client.

shu-g: Ahh, I see, when we see a new chunk and `append_chunk_to_file()`, we need the filename.
I agree, for now can forward filename with write requests.

shu-g (approved):
LGTM

src/sadmd/sadmd.cpp
```diff
@@ -383,16 +414,16 @@ reintroduce_chunks_to_network(util::file_chunks ids)
 	}
 }
 
-void sadmd::
+bool sadmd::
 add_chunk_to_server(chunkid cid, version v, serverid sid)
 {
-	if (!is_active(sid)) return;	
-	auto server_ptr = &(chunk_server_metadata_.at(sid));
+	if (!is_active(sid)) return false;
```
damanmkumar: Is this check necessary?

src/sadcd/sadcd.cpp
```diff
@@ -112,4 +112,21 @@ join_network()
 
 }
 
+bool sadcd::
+notify_master_of_write(chunkid chunk, version version_num, std::string const& filename)
+{
+	auto ch = msgs::channel{master_.connect()};
```
damanmkumar: We need to check if `connect()` returns a valid `socket`

src/sadcd/sadcd.cpp
```diff
@@ -112,4 +112,21 @@ join_network()
...
+
+	auto cwn = msgs::master::chunk_write_notification
+	{
+		serverid_,
+		chunk,
+		version_num,
+		filename
+	};
+
+	msgs::master::serializer{}.serialize(cwn, ch);
```
damanmkumar, line 147: We need confirmation from the master that the write went through. Otherwise, we can't say to the client that the write went through. A generic `acknowledgement` will do. Let's implement that in another PR. We should definitely add a TODO.

include/sadfs/sadmd/sadmd.hpp
```diff
@@ -44,7 +43,8 @@ struct chunk_info
 {
 	version latest_version;
 	// all the places this chunk is stored and what version is there
-	std::vector<std::pair<chunk_server_info*, version> > locations;
+	std::unordered_map<serverid, version> locations;
+	//std::vector<std::pair<chunk_server_info*, version> > locations;
```
damanmkumar: Could you also remove this commented code?

src/sadmd/sadmd.cpp
```diff
@@ -208,6 +205,40 @@ handle(msgs::master::join_network_request const& jnr, msgs::channel const& ch)
 	return true;
 }
 
+bool sadmd::
+handle(msgs::master::chunk_write_notification const& cwn, msgs::channel const& ch)
+{
+	auto chunk = cwn.chunk_id();
+	auto server = cwn.server_id();
+
+	if (chunk_metadata_.count(chunk))
```
damanmkumar: Can we avoid this redundant lookup?

```
auto it = chunk_metadata_.end();
if ((it = chunk_metadata_.find(...)) != chunk_metadata_.end())
{
    auto& info = it->second;
    ...
```

Same with `info.locations` as well

damanmkumar (changes requested):
A couple of minor comments

damanmkumar:
Looks like there are conflicts. Maybe dev needs to be merged in.

src/sadmd/sadmd.cpp
abmono93: Without it you could add a serverid to a chunk's location list that isn't also in `chunk_server_metadata_`. Later if that chunk's locations are requested, the master will try to look up that server's info in using that serverid in `chunk_server_metadata_`, causing a seg fault

damanmkumar: There wouldn't be a seg fault, but the server would just turn out to be unresponsive, right? The `server->valid_until > time::now()` check would fail.

But if you mean that we also need to make sure that this server is on the network and responsive, then this check needs to happen as the first step in processing this `chunk_write_notify` request, correct?

abmono93: The seg fault would occur when trying to look up the `valid_until` field for a server with that `serverid`

abmono93: > But if you mean that we also need to make sure that this server is on the network and responsive, then this check needs to happen as the first step in processing this chunk_write_notify request, correct?

Let me test this out. The tests I did the check in `is_active` caught all these situations

abmono93: Sorry you were right, it's not a seg fault exactly:
`libc++abi.dylib: terminating with uncaught exception of type std::out_of_range: unordered_map::at: key not found
Abort trap: 6`

But to avoid this the check in `add_chunk_to_server` is sufficient

damanmkumar: Which line throws this exception? We're looking up an id that does not exist in a map, which is the source of that error.

The reason I'm not convinced we should be performing this check is that we should be checking if a server is active only when we have reason to believe that the server might be down. In this case, since the server has just told us that it has a new version of a chunk, it should be valid.

I have a feeling that it is the
```
if (!chunk_server_metadata_.count(id))
```
in `is_active` that is masking some other issue.

PR #81, opened 2020-12-10
made socket class throw fewer exceptions

damanmkumar:
- made service::connect noexcept
- sadcd::join_network doesn't need to catch exceptions anymore

### DIFF
--- a/src/sadcd/sadcd.cpp
+++ b/src/sadcd/sadcd.cpp
@@ -107,28 +107,20 @@ sadcd::start_main(std::promise<void>       death,
 bool
 sadcd::join_network()
 {
-    /*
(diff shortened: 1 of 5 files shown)


src/sadcd/sadcd.cpp
```diff
@@ -107,28 +107,20 @@ sadcd::start_main(std::promise<void>       death,
 bool
 sadcd::join_network()
 {
-    /*
-    //auto ch = channel{socket{domain::inet, type::stream, -1}}; // dummy
-    channel try
+    auto sock = master_.connect();
+    if (!sock.valid())
```
abmono93, line 111: This is a nice solution

abmono93 (approved):

PR #84, opened 2020-12-10
 implemented acknowledgement message for chunk servers

damanmkumar:
- templated class that is yet to be specialized for
  clients
- removed log function overloads that took std::string


abmono93 (approved):
LGTM

PR #88, opened 2020-12-11
implemented control message header containing host-id

damanmkumar:
- sadmd is untested, but there should be no surprises based on the following results
- dummy-sadmd is able to read and print dummy-sadcd's
  host id
- dummy-sadmd is able to read and print the real sadcd's
  host id
- lots of changes from clang-format
**NOTE:** important files are
- msgs/master/message_processor.hpp
- serializer_impl.hpp
- sadcd.hpp & sadcd.cpp
- sadmd.hpp & sadmd.cpp
- heart.hpp & heart.cpp


abmono93 (approved):
LGTM

PR #92, opened 2020-12-11
changed chunk request to stream request

damanmkumar:
- implemented data streaming capabilities
- sadcd doesn't yet handle these
- fixed a bug in the default constructor of `socket`


example/stream-req.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+using namespace sadfs;
+constexpr auto delim = std::string_view{"******************************\n"};
+
+void
+print_stream_req(msgs::chunk::stream_request const& req)
+{
+    std::cout << delim << "Stream Request:"
+              << "\nChunk ID:  " << to_string(req.chunk_id())
+              << "\nOffset:    " << req.offset()
```
shu-g: I think we should add version_num to the request after chunk_id.

shu-g (commented):
Are you actually implementing stream? It seems misleading otherwise.
I was thinking read_chunk_request (think this is more consistent) with this api (with version_num) should be good.
For writes, write_chunk_request should also send filename(so that it can notify master) and other chunk locations. First chunk server can then call write_chunk_request with all fields as write_chunk_request except locations field empty.

damanmkumar:
> Are you actually implementing stream? It seems misleading otherwise.

Sending data over for writes, and receiving it for reads is streaming, right?

> I was thinking read_chunk_request (think this is more consistent) with this api (with version_num) should be good.

This actually makes me realize that `std::string&& data` should be an optional parameter. Also, why would a client need to request a specific version of a chunk (for reads or writes)?

> For writes, write_chunk_request should also send filename(so that it can notify master) and other chunk locations. First chunk server can then call write_chunk_request with all fields as write_chunk_request except locations field empty.

Since we aren't going to get this integrated before our meeting today, I figured we can ignore the hack we've put together where chunk servers are sent the filename.

Your comment reminds me -- we're missing the list of servers as well. This combined with the fact that we provide data to the request only while writing makes me think that separate read/write requests are cleaner. 

So, I'm convinced that separate read and write requests are the way to go. But, I don't see the point of a version number field.

Tagging you so get a notification: @shu-g

shu-g:
I thought streaming meant maintaining a persistent connection and exchanging data at the application level.

It is possible for write to be happening at version v4, but since that is still in progress, master returns v3 in chunk_location_request which client can simply forward to chunk_server. 
Also, it gives us extensibility (don't think this is right word) for features such as snapshots (I'm also laughing) without added effort.
@damanmkumar

abmono93:
> Also, why would a client need to request a specific version of a chunk (for reads or writes)

@damanmkumar on a write the client has to include the latest version number so the chunk server can update the master with correct info. Chunk servers may have missed intermediate versions so even if they know what version they used to have, they still need this info from the master (via the client). So they need to send a chunk write notification to the master saying they now have `version_num they received` + 1, for versioning to work. The client can just be a pass through.

damanmkumar:
> > Also, why would a client need to request a specific version of a chunk (for reads or writes)
> 
> @damanmkumar on a write the client has to include the latest version number so the chunk server can update the master with correct info. Chunk servers may have missed intermediate versions so even if they know what version they used to have, they still need this info from the master (via the client).

The master only ever gives a client the IP + port of chunk servers that have the latest version. Since clients don't cache the entire chunk at their end, writes and appends are only ever going to be for specific sections of chunks (append these 512 bytes, or write these 1024 bytes at offset 7324). So, a server that is behind can only "catch up" if another chunk server gives it a copy of the latest chunk.

>  So they need to send a chunk write notification to the master saying they now have `version_num they received` + 1, for versioning to work. The client can just be a pass through.

Since clients don't overwrite entire chunks, versions can only be bumped up from their current version by 1. Except when it receives a new chunk from a sibling chunk server.

EDIT: Ugh, forgot to tag you, @abmono93

include/sadfs/msgs/chunk/messages.hpp
```diff
@@ -22,44 +22,88 @@ enum class msg_type
 {
     unknown,
     acknowledgement,
-    chunk_request,
+    read_request,
+    read_response,
```
shu-g: Shouldn't this be in client/messages.hpp?
Noticed this when merged these changes into my branch.

abmono93:
> The master only ever gives a client the IP + port of chunk servers that have the latest version.

This is only true for read requests. Chunk location requests for a write as currently handled by master will give all locations and what the latest version is. Some of those locations may have older versions. 

> Since clients don't cache the entire chunk at their end

I've been operating under the assumption that they do cache the chunk and stream their new copy of the whole chunk on each write. But that may be wrong.

@damanmkumar

damanmkumar:
> Chunk location requests for a write as currently handled by master will give all locations and what the latest version is. Some of those locations may have older versions.

We'll have to fix this. `fuse` hands us read/write syscalls and those are only for specific sections of the file. If we were to cache chunks, we would have to do quite a bit of bookkeeping to match syscalls to cached chunks.

@abmono93

include/sadfs/msgs/chunk/messages.hpp
abmono93: chunk/messages.hpp is all the messages that a chunk can receive, regardless of who sends them. So `read_response` should be in client, but `read_request` should be here.

damanmkumar:
Beyond help

PR #93, opened 2020-12-11
Release lock

abmono93:
This is the simplest possible working version of a release_lock message and handler. The master currently has no mechanism to verify that the sender holds the lock, so it simply trusts any incoming message is from the right client and unlocks the file. It also doesn't send an ack yet.

### DIFF
--- a/include/sadfs/msgs/master/messages.hpp
+++ b/include/sadfs/msgs/master/messages.hpp
@@ -29,7 +29,7 @@ enum class msg_type
     chunk_location_request,
     chunk_server_heartbeat,
     join_network_request,
-    release_lock
+    release_lock,
 };
 
 class file_metadata_request
--- a/include/sadfs/msgs/master/messages.hpp
+++ b/include/sadfs/msgs/master/messages.hpp
@@ -28,7 +28,8 @@ enum class msg_type
     chunk_write_notification,
     chunk_location_request,
     chunk_server_heartbeat,
-    join_network_request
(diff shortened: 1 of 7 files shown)


include/sadfs/msgs/master/messages.hpp
```diff
@@ -28,7 +28,8 @@ enum class msg_type
     chunk_write_notification,
     chunk_location_request,
     chunk_server_heartbeat,
-    join_network_request
+    join_network_request,
+    release_lock
```
damanmkumar: Nit: We can add a trailing comma at the end of the last `enum`as well, which will not show a two line diff like here. Also, merges are more likely to succeed when multiple people edit files.

src/proto/master.proto
```diff
@@ -42,10 +42,16 @@ message file_metadata_request {
...
+// notification of release of write lock
+message release_lock {
+	string filename	= 1;
+}
+
 // container for control messages
 message message_container {
 	sadfs.proto.header  header = 1;
 	oneof               msg {
+		release_lock release_lock = 5;
```
damanmkumar, line 54: Nit: formatting is off

damanmkumar (approved):
Couple of minor comments that can wait. Looks good otherwise

src/proto/master.proto
abmono93: Huh. Does format.sh not format this type of thing?

PR #94, opened 2020-12-11
Implement sadfilesys::read()

shu-g:
* Move bytes_per_chunk to sadfs/constants.hpp
* Implement read()
* Modify assigned errors


abmono93 (approved):
Looks great

include/sadfs/constants.hpp
```diff
@@ -0,0 +1,11 @@
+#ifndef SADFS_CONSTANTS_HPP
+#define SADFS_CONSTANTS_HPP
+
+namespace sadfs
+{
+
+constexpr auto bytes_per_chunk = 64 * 1024 * 1024; // 64MB
```
damanmkumar, line 9: Let's add this to the `constants` namespace

src/sadcd/heart.cpp
```diff
@@ -23,7 +23,7 @@ heart::beat(std::promise<void> death, std::future<void> stop_token)
 
     // send heartbeats until requested to stop
     using namespace std::chrono_literals;
-    while (stop_token.wait_for(1s) != std::future_status::ready)
+    while (stop_token.wait_for(15s) != std::future_status::ready)
```
damanmkumar, line 26: Isn't 15 seconds too long a time duration? 1 second is pretty long itself, especially when we're talking about a system like this. But, anything over 5 seconds is way too long, IMO

src/sadcd/sadcd.cpp
```diff
@@ -97,13 +98,52 @@ void
...
+{
+	auto response = msgs::client::read_response{
+		true,		/* ok */
+		"chunk chonk"	/* data */
+	};
+	auto result = msgs::client::serializer{}.serialize(response, ch);
+	ch.flush();
+	return result;
+}
+
```
damanmkumar: #98 implements this functionality. 
You might need a local copy of this for testing on your end.

src/sadfsd/sadfilesys.cpp
```diff
@@ -56,8 +60,17 @@ int
 sadfilesys::getattr(char const* path, struct stat* stbuf)
 {
     logger::debug("gettattr() called at " + std::string(path));
-
     auto result{0};
+
+    if ((strlen(path) >= 2 && path[0] == '/' && path[1] == '.') ||
+    	strcmp(path, "/autorun.inf") == 0)
```
damanmkumar, line 66: What are `/.` and `/autorun.inf` ?

src/sadfsd/sadfilesys.cpp
```diff
@@ -56,8 +60,17 @@ int
...
+    {
+        result = -ENOENT; 	// No such file or directory
+        logger::debug("gettattr() failed with error " +
+                      std::string(strerror(-result)));
```
damanmkumar, line 70: This reads like a message logged by someone that calls `getattt`. Our debug statement should help us figure out what sadfilesys is doing, which is rejecting this request because it is for `/.` or `/autorun.inf`

damanmkumar (changes requested):
Left a couple of comments a while ago

include/sadfs/constants.hpp
shu-g: Oh right.

src/sadcd/heart.cpp
shu-g: I modified it to 15s to make it easier for me to parse logs during testing.
I would think in an actual system, it should be on the order of minutes since it is unlikely that multiple servers fail within that span and the request can still be served to client through the other chunk servers.
Also heartbeat every second from 10-50 chunk servers is unnecessarily creating network traffic and load at the master server. I don't see the requirement for strict consistency in our use case between master and chunk server.

src/sadcd/sadcd.cpp
shu-g: Yep, will keep a copy.

src/sadfsd/sadfilesys.cpp
shu-g: When the filesystem is mounted, I observe requests by the OS to a couple of hidden files like /.Trash, /.Trash1000 and to the file /autorun.inf. We should be able to remove them when we start testing with actual sadmd, we'll get not-ok from the master server and return error. At the moment using dummy-sadmd, we return ok for every request and then the os tries to read these files that don't exist.

src/sadcd/heart.cpp
damanmkumar: I see what you mean, but shouldn't our number one priority be fast access for clients of the FS? If a popular server goes down, and we serve its IP to all those clients for a few minutes, it would be a disaster (hyperbole, definitely). I think it is realistic for a master server to be serving hundreds of thousands of requests per second at a minimum, and a few hundred lightweight heartbeat requests should be insignificant.

PR #97, opened 2020-12-12
Miscellaneous changes

abmono93:
- Created a simple test so I can quickly check if I broke something in sadmd. We can also remove send-req.cpp unless someone else has a use for it
- Changed protobufs that include a serverid to use the header instead
- We now return only servers with the latest verion on both reads and writes
- Added some logging
- Removed a check in append_chunk_to_file which was redundant with validation in the handler for chunk write notifications


example/CMakeLists.txt
```diff
@@ -11,3 +11,4 @@ add_executable (send-req 	send-req.cpp)
 add_executable (dummy-sadcd 	dummy-sadcd.cpp)
 add_executable (open-file 	open-file.cpp)
 add_executable (chunk-read-req  chunk-read-req.cpp)
+add_executable (tests 	tests.cpp)
```
damanmkumar: How about sadmd-tests?

abmono93: Yea I am totally ok with that. But we could also add tests of the other components too. Actually probably only sadcd would be feasible... Up to you guys if you think you'll use it.

damanmkumar: What I thought of was `sadmd-tests` for testing `sadmd`, and `sadcd-test` for `sadcd`. Also, you're right, I don't see how we could test `sadfsd` this way

src/sadmd/sadmd.cpp
```diff
@@ -275,21 +275,31 @@ sadmd::handle(msgs::master::chunk_location_request const& clr,
...
 {
-    auto server = cwn.server_id();
-    // validate request - TODO: Add error logging
-    if (!is_active(server) || !files_.count(cwn.filename()))
+    auto server = header.host_id;
+    // validate request
+    if (!is_active(server))
+    {
+        std::cerr
+            << "Error: write notification from server not on the network\n";
```
damanmkumar: This check would also fail if the server hasn't been sending heartbeats, right? How about `"server not on network, or not sending heartbeats"`?

Also, let's use `logger::error` since you're touching this anyway.

src/sadmd/sadmd.cpp
```diff
@@ -275,21 +275,31 @@ sadmd::handle(msgs::master::chunk_location_request const& clr,
...
         return false;
```
damanmkumar: `handle(...)` should return `true` if it could "handle" a message. If we return `false` here, `processor.process_next(...)` will return `false` as well, giving the impression that something went wrong while processing the request.

A valid scenario where we would return `false` is when sadcd fails to read from disk, which would prevent it from handling a request. I realised this while typing this comment out, so, you might want to see this new change in `sadcd.cpp` as well.

Ugh, this means that we'll have to take a look at the return values of all handle functions.

src/sadmd/sadmd.cpp
```diff
@@ -334,6 +344,9 @@ sadmd::handle(msgs::master::chunk_write_notification const& cwn,
         {
             file_info_.size = ((num_chunks - 1) * constants::bytes_per_chunk) +
                               cwn.new_size();
+            logger::debug(std::string_view{"Updated size of " +
```
damanmkumar: `logger` functions accept `std::string`s as well, right? Or did I remove those methods? I remember removing them, but thought I added them back in

shu-g (approved):
LGTM

example/CMakeLists.txt
abmono93: Yea actually that makes a lot more sense. Otherwise you need to be running two servers to test

src/sadmd/sadmd.cpp
abmono93: Not sure. I use string_view because the compiler told me to once when I tried using a string and I haven't tried again since

damanmkumar (approved):
LGTM

PR #98, opened 2020-12-12
 implemented chunk read

damanmkumar:
- chunk server does not persist data to disk, chunk info is
  hardcoded for testing
- added a create-chunk program that creates a 64 MB chunk
  with the format:
```
  <36-char UUID>00000...\n
  <36-char UUID>11111...\n
  ...
  <36-char UUID>99999...\n
  <36-char UUID>00000...\n
  ...
```
- create-chunks.sh creates the chunks whose info is hardcoded
- chunk-read-req send a few requests, some of which are rejected,
  and some of which retrieve data. Only data transfer of 64 bytes
  tested.


src/sadcd/sadcd.cpp
```diff
@@ -145,4 +196,103 @@ sadcd::notify_master_of_write(chunkid chunk, version version_num,
...
+        };
+        return msgs::client::serializer{{serverid_}}.serialize(
+                   msgs::client::read_response{ok, std::move(data)}, ch) &&
+               flush();
+    };
+    // validate chunk_id
+    auto it = chunk_metadata_.find(req.chunk_id());
+    if (it == chunk_metadata_.end())
+    {
+        logger::debug("chunk does exist on this server"sv);
```
abmono93: Did you mean "does not exist" ?

src/sadcd/sadcd.cpp
```diff
@@ -145,4 +196,103 @@ sadcd::notify_master_of_write(chunkid chunk, version version_num,
...
 
+// ==================================================================
+//                           request_handler
+// ==================================================================
+namespace
+{
+
+// get version info of the newest version of a chunk
+auto latest_version = [](auto it) {
+    return std::max_element(it->second.versions().begin(),
```
abmono93, line 207: Are we having the chunk server store multiple versions?

abmono93 (approved):
One minor thing, one question, but looks good.

src/sadcd/sadcd.cpp
damanmkumar: Yup! Good catch, thanks!

src/sadcd/sadcd.cpp
```diff
@@ -145,4 +196,105 @@ sadcd::notify_master_of_write(chunkid chunk, version version_num,
...
+    }
+
+    file.seekg(offset, std::ios_base::beg);
+    file.read(buf.data(), length);
+
+    if (file.good() && file.gcount() == length)
+    {
+        return true;
+    }
+    logger::debug("file read error"sv);
```
shu-g, line 239: Even when the client program requests to read 10 bytes, we get a request on the fuse side of 8191 bytes. Chunk should return the min of that chunk size and requested read size.

damanmkumar: Won't that be handled by the client code? If `sadfsd` requests 8191 bytes, `sadcd` will return 8191 bytes.

src/sadcd/sadcd.cpp
```diff
@@ -145,4 +196,105 @@ sadcd::notify_master_of_write(chunkid chunk, version version_num,
...
+    {
+        logger::debug("chunk does not exist on this server"sv);
+        return respond(false, {});
+    }
+
+    // validate offset and length
+    auto m_it = latest_version(it);
+    if (req.offset() >= m_it->size() ||
+        (m_it->size() - req.offset()) < req.length())
+    {
```
shu-g, line 286: We should remove this check.

src/sadcd/sadcd.cpp
damanmkumar: No, not for now. We simply bump up the version and size on appends.

shu-g (changes requested):
Pointed issue with check size of requested chunk, other than that looks good, we should merge this before we merge read() PR, there is a bit of overlap that I should resolve once I pull these.

src/sadcd/sadcd.cpp
damanmkumar: File read is going to fail if we remove this check, and a client requests more data than there is in a chunk. If a file only has 10 bytes, and fuse asks for 8191 bytes, shouldn't we read 10 bytes into the buffer, and return 10?

shu-g (approved):
The check can be performed at the sadfilesys end, ignore previous comments.

PR #99, opened 2020-12-13
Remove depracated declaration from sadcd.hpp

shu-g:
* Remove handle() declaration from sadcd.hpp
* Comment out test code from sadcd.cpp


damanmkumar (approved):
LGTM!

PR #100, opened 2020-12-13
Sadmd handles heartbeats, sends acks

abmono93:


src/sadmd/sadmd.cpp
```diff
@@ -181,9 +182,11 @@ bool
 sadmd::handle(msgs::master::create_file_request const& cfr,
               msgs::message_header const&, msgs::channel const& ch)
 {
-    create_file(cfr.filename());
-    // TODO: send ack if create_file succeeded
-    return true;
+    auto ok     = create_file(cfr.filename());
+    auto ack    = msgs::client::acknowledgement{ok};
+    auto result = msgs::client::serializer{}.serialize(ack, ch);
+    ch.flush();
```
damanmkumar: Explicit flushes aren't necessary before returning since the output buffers will be flushed as soon as the channel object's destructor is called

shu-g (approved):
LGTM

damanmkumar (approved):
LGTM

PR #101, opened 2020-12-14
implemented append_request, and corresponding chunk server-side implementation

damanmkumar:
- added a few tests to verify basic functionality
- chunk replication is pending


src/sadcd/sadcd.cpp
```diff
@@ -203,26 +183,47 @@ namespace
...
+                            it->second.mutable_versions()->end(),
                             [](auto lmeta, auto rmeta) {
                                 return lmeta.version() < rmeta.version();
                             });
 };
 
+auto is_valid = [](msgs::chunk::append_request const& req, auto it,
+                   bool new_chunk) {
+    // validate length
+    if (req.length() > 64 * 1024 * 1024)
```
shu-g, line 196: We can use `constants::bytes_per_chunk` from <sadfs/constants.hpp>

src/sadcd/sadcd.cpp
```diff
@@ -203,26 +183,47 @@ namespace
...
+    {
+        logger::error("invalid length requested"sv);
+        return false;
+    }
+    // existing chunk
+    if (!new_chunk)
+    {
+        auto m_it = latest_version(it);
+        if ((64 * 1024 * 1024 - m_it->size()) < req.length())
```
shu-g, line 205: `constants::bytes_per_chunk`

src/sadcd/io.cpp
```diff
@@ -0,0 +1,113 @@
...
+    {
+        return false;
+    }
+
+    auto ch = msgs::channel{std::move(sock)};
+
+    auto cwn = msgs::master::chunk_write_notification{
+        spec.id, spec.ver + 1, spec.filename, spec.size + spec.length};
+
+    // TODO: confirm from master that the write went through
```
abmono93, line 91: I implemented this TODO in `sadcd::notify_master_of_write` in my PR. I can move it here after this is merged.

abmono93 (approved):

src/sadcd/sadcd.cpp
```diff
@@ -296,4 +298,92 @@ request_handler::handle(msgs::chunk::read_request const& req,
...
+    }();
+    entry->set_version(entry->version() + 1);
+    entry->set_size(entry->size() + req.length());
+
+    // TODO: forward appends to replications servers
+    return true;
+}
+
+bool
+request_handler::ack_client(bool ok, msgs::channel const& ch)
```
abmono93, line 378: I am going to put something similar in sadmd. It doesn't need to be a member function though, does it?

damanmkumar: Not, really, no. I didn't want to pass in serverid as an argument every time -- so, I made it a member function. But now that I think about it, we don't need to send our serverid at this point, since the client knows who it established a connection with. So, yeah, we can totally make it a common utility function.

src/sadcd/io.cpp
damanmkumar: Yup, I saw that, thanks!

src/sadcd/sadcd.cpp
abmono93: Oh right didn't notice the `serverid_` in there. Also, the client doesn't even know the server's id, only the service it listens on. So it couldn't check that the ack's `host_id` matches even if it wanted to.

src/sadcd/sadcd.cpp
```diff
@@ -296,4 +298,92 @@ request_handler::handle(msgs::chunk::read_request const& req,
...
+        return false;
+    }
+
+    // update data structures in memory
+    auto entry = [this, new_chunk, it, &req]() {
+        if (new_chunk)
+        {
+            return chunk_metadata_[req.chunk_id()].add_versions();
+        }
+        return &*latest_version(it);
```
shu-g, line 368: What is happening here? Why do we return reference to something we dereference?

shu-g (commented):
Nice changes!
Didn't we decide to have files with version num inside the the uuid folder? Are we overwriting the current file and just updating the metadata due to time constraints or there is a design change?
Also, I still feel forwarding the version num from the client that it gets from master is more intuitive, the chunk just updates it version num to (v+1).

src/sadcd/sadcd.cpp
damanmkumar: Yup, you're right. But, in case the client wanted to make sure it's a "known" ID, maybe? I wouldn't know where to start looking if we spoke about security, though, haha. So, yeah, it's totally irrelevant, and a waste of network bandwidth, maybe

src/sadcd/sadcd.cpp
damanmkumar: `latest_version` returns an iterator type, whereas `add_versions()` returns a `proto::service*` type, and the only way I could get the return types to match was this way. Dereferencing the iterator will give us a reference to `proto::service`, the address of which I return by applying the `&` operator. It's ugly, but correct and doesn't have the possibility of memory leaks and whatnot (at least as far as I can think of). 

We should definitely "fix" it once we've got an MVP  going.

src/sadcd/sadcd.cpp
damanmkumar: Yup, I had a comment in there, but it looks like it slipped my mind. Also, is `chunk_size` smaller/cleaner?

damanmkumar:
> Nice changes!
> Didn't we decide to have files with version num inside the the uuid folder? Are we overwriting the current file and just updating the metadata due to time constraints or there is a design change?

A bit of both, actually, but mainly due to time constraints. I couldn't figure out a clean way to avoid an edge case where we tell the master server we have version V+1, but the rename fails, leaving us in an inconsistent state. We should definitely come up with a nicer strategy. Also, I realised that storing every version (like a copy-on-write operation) would kill the performance of the operations (unless a snapshot is requested).

> Also, I still feel forwarding the version num from the client that it gets from master is more intuitive, the chunk just updates it version num to (v+1).

But, I don't see the use case of writing to a chunk that *isn't* the latest version. Also, aren't chunk servers the source of truth for version numbers?

src/sadcd/sadcd.cpp
shu-g: Yeah, I agree the name could be better. How about `chunk_capacity`? Makes it harder to get it confused with the actual size of a chunk.

src/sadcd/sadcd.cpp
shu-g: Ahh I see, interesting!

PR #103, opened 2020-12-14
implemented chunk append request forwarding

damanmkumar:
- chunk servers now forward all append requests to 'downstream'
  servers by chaining requests: A --forward--> B --forward--> C
- if a server in the middle of a 'chain' is unreachable, servers
  'get around' the unreachable server, and keep trying to reach
  the next downstream server, and stop when they find one. It is
  then the responsibility of this server to pass the append/write
  downstream

made versioning simpler

- all sadcd-tests pass, with forwarding to one other server
- bytes_per_chunk -> chunk_capacity
- replaced 64 * 1024 * 1024 with constants::chunk_capacity


shu-g (approved):
LGTM!

PR #104, opened 2020-12-14
Fuse write

shu-g:
Implemented the sadfilesys::write() function.
Yet to test on writes spanning multiple chunks and verify if any issues on macOS.
To "test"
- write I used `echo "chunk chonk" > path/to/mount/filename`
- read I used `cat path/to/mount/filename`
- append, waited 1 minute for ttl to run out, then `echo "chank chenk" >> path/to/mount/filename`

### DIFF
--- a/src/sadfsd/sadfilesys.cpp
+++ b/src/sadfsd/sadfilesys.cpp
@@ -51,7 +51,8 @@ sadfilesys::load_operations()
     };
     operations_.read = [](char const* path, char* buf, size_t size,
                           off_t offset, fuse_file_info* fi) -> int {
-        return this_()->read(path, buf, size, offset, fi);
+        return this_()->read(path, buf, static_cast<uint32_t>(size),
+			     static_cast<uint32_t>(offset), fi);
     };
     operations_.readdir = [](char const* path, void* buf,
                              fuse_fill_dir_t filler, off_t off,
@@ -60,7 +61,8 @@ sadfilesys::load_operations()
     };
     operations_.write = [](char const* path, const char* buf, size_t size,
                           off_t offset, fuse_file_info* fi) -> int {
-        return this_()->write(path, buf, size, offset, fi);
+        return this_()->write(path, buf, static_cast<uint32_t>(size),
+			      static_cast<uint32_t>(offset), fi);
     };
 }
 
@@ -189,7 +191,7 @@ sadfilesys::open(char const* path, fuse_file_info* fi)
 }
 
 int
-sadfilesys::read(char const* path, char* buf, size_t size, off_t offset,
+sadfilesys::read(char const* path, char* buf, uint32_t size, uint32_t offset,
                  fuse_file_info* fi)
 {
     logger::debug("read() called at " + std::string(path) + " to read " +
@@ -201,7 +203,7 @@ sadfilesys::read(char const* path, char* buf, size_t size, off_t offset,
     {
     	msgs::io_type::read,
 	std::string{path},
-	static_cast<uint32_t> (offset / constants::bytes_per_chunk)
+	offset / constants::bytes_per_chunk
     };
     auto location_response	= msgs::client::chunk_location_response{};
     auto master_serializer 	= msgs::master::serializer{};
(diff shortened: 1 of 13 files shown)


src/sadfsd/sadfilesys.cpp
```diff
@@ -47,13 +51,60 @@ sadfilesys::load_operations()
...
+		      std::string(strerror(-result)));
+	return result;
+    }
+    auto ch = msgs::channel{std::move(sock)};
+    auto flush = [](auto const& ch) {
+    	ch.flush();
+	return true;
+    };
+
+    if (!(serializer.serialize(request, ch) && flush(ch)))
```
damanmkumar, line 98: The master server now sends acks, so, we should verify that before saying that file creation succeeded

src/sadfsd/sadfilesys.cpp
```diff
@@ -282,4 +329,158 @@ sadfilesys::read(char const* path, char* buf, size_t size, off_t offset,
...
+    }
+    if (location_response.locations_size() == 0)
+    {
+    	result =  -EPROTO;	// Protocol error
+        logger::debug("write() empty chunk location returned " +
+                      std::string(strerror(-result)));
+        return result;
+    }
+
+    auto file_size	= location_response.file_size();
```
abmono93, line 440: I thought file size should come from file_metadata. Where do you tell the master to create file? The flow I had in mind was something like:
- get ok and size  from file_metadata_request(filename)
- if !ok create_file(filename)
- figure out which chunk you need based on size
- send chunk_location_request

The way this is written I think sadmd will reject all writes since the file will never exist

src/sadfsd/sadfilesys.cpp
```diff
@@ -204,20 +255,16 @@ sadfilesys::read(char const* path, char* buf, size_t size, off_t offset,
     
     auto file_size	= location_response.file_size();
     logger::debug("read() got file size " + std::to_string(file_size));
-    if (static_cast<uint32_t>(offset) >= file_size)
+    if (offset >= file_size)
     {
     	return result;
```
damanmkumar, line 307: Should we return `0` when `offset` is invalid?

src/sadfsd/sadfilesys.cpp
damanmkumar: It looks like FUSE has a `create` "syscall" as well. I see a `create_file_request` there.
Also, why should we try to create a file in `write`?

damanmkumar: Given the API and the fact that we saw different file descriptors after opening files, I get the feeling that FUSE is handling things such as calling create before write, if the application code is something like:
```
int fd = open("new-file.dat", O_WRONLY | O_CREAT);
write(fd, buf, size);
```

abmono93: > Also, why should we try to create a file in write?

I was thinking you'd check for the O_CREAT flag. But looking at the write(2) man page that's not actually a flag, so I'm getting mixed up I guess. Anyway if fuse calls `create` for us this is a moot point. Still I don't see a reason to report the file size in both file_metadata_response and chunk_ location_response, but I guess it doesn't hurt

damanmkumar: I was just wondering if there is any benefit in continuing to do this, since we get all our information in a single request. And then it got me thinking about whether FUSE somehow caches the file size, in whichh case, we will have troubles later on. What if FUSE thinks a file is X bytes, but someone else appended to the file and it is now Y bytes? In this case, FUSE will want to write at X bytes (thinking it is appending to the file). I guess we don't have to think of these things now.

src/sadfsd/sadfilesys.cpp
```diff
@@ -282,4 +329,158 @@ sadfilesys::read(char const* path, char* buf, size_t size, off_t offset,
...
+     	auto ret = write(path, buf + write_size, size - write_size,
+		        offset + write_size, fi);
+	if (ret < 0)
+	{
+		return result;
+	}
+	return result + ret;
+     }
+     
+     logger::debug("write() finished"sv);
```
abmono93, line 547: I don't think we need to wait out the ttl, why not send a `release_lock` ?

src/sadfsd/sadfilesys.cpp
shu-g: Oh ok.

PR #105, opened 2020-12-15
Fuse write and readdir

shu-g:
* Implemented write() and readdir()
* Rebased on dev thus incorporating the replication changes
* Yet to be tested on macOS

### DIFF
--- a/src/sadfsd/sadfilesys.cpp
+++ b/src/sadfsd/sadfilesys.cpp
@@ -51,7 +51,8 @@ sadfilesys::load_operations()
     };
     operations_.read = [](char const* path, char* buf, size_t size,
                           off_t offset, fuse_file_info* fi) -> int {
-        return this_()->read(path, buf, size, offset, fi);
+        return this_()->read(path, buf, static_cast<uint32_t>(size),
+			     static_cast<uint32_t>(offset), fi);
     };
     operations_.readdir = [](char const* path, void* buf,
                              fuse_fill_dir_t filler, off_t off,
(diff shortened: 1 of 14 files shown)


src/sadfsd/sadfilesys.cpp
```diff
@@ -75,8 +126,9 @@ sadfilesys::getattr(char const* path, struct stat* stbuf)
     if (strcmp(path, "/") == 0)
     {
         logger::debug("path identified as base directory"sv);
-        stbuf->st_mode = S_IFDIR | 0755;
+        stbuf->st_mode = S_IFDIR | 0777;
```
damanmkumar, line 129: Why does 755 not work?

src/sadfsd/sadfilesys.cpp
```diff
@@ -127,19 +180,65 @@ int
...
+    {
+    	result = -EPROTONOSUPPORT; 	// Protocol not supported
+	logger::debug("create() failed with error " +
+		      std::string(strerror(-result)));
+	return result;
+    }
+    
+    for (auto i = 0; i < response.filenames_size(); ++i)
+    {
+    	filler(buf, response.filename(i).c_str() + 1, NULL, 0);
```
damanmkumar, line 227: Let's qualify the names of functions we haven't written like `::functionname` so that it is clear that it is an external function. Same goes for `strcpy`, `strcmp`, and the like, which are in the `std` namespace.

src/sadfsd/sadfilesys.cpp
```diff
@@ -199,7 +298,7 @@ sadfilesys::read(char const* path, char* buf, size_t size, off_t offset,
 
     auto file_size = location_response.file_size();
     logger::debug("read() got file size " + std::to_string(file_size));
-    if (static_cast<uint32_t>(offset) >= file_size)
+    if (offset >= file_size)
     {
         return result;
```
damanmkumar, line 303: Should this not be an error?

